PR #4, opened 2021-05-29
Operator support for IxiaTG

hashwini-keysight:
Hi Marcus,

We have added operator support for IxiaTG. We withdrew previous PRs for minor change.

Thanks,
Himanshu


cmd/root.go
```diff
@@ -113,11 +113,13 @@ func createFn(cmd *cobra.Command, args []string) error {
 	if err := t.Push(cmd.Context()); err != nil {
 		return err
 	}
-	fmt.Fprintf(out, "Topology %q created", topopb.Name)
+	fmt.Fprintf(out, "Topology %q created\n", topopb.Name)
 	r, err := t.Resources(cmd.Context())
-	fmt.Fprintf(out, "Pods:\n")
-	for _, p := range r.Pods {
-		fmt.Fprintf(out, "%s\n", p.Name)
+	if r != nil {
```
marcushines, line 118: there just needs to be an err check here - r should never be nil

topo/topo.go
```diff
@@ -210,10 +212,13 @@ func (m *Manager) Push(ctx context.Context) error {
 		if err := n.CreateService(ctx); err != nil {
 			return err
 		}
-		if err := n.CreatePod(ctx); err != nil {
```
marcushines, line 213: you are breaking all other implementations if you do this
if you are going to suggest this change then also update all implementations in the same PR

marcushines (commented):
The original thought behind the struct hierarchy is that the node's don't need to know about k8s - you are fundamentally shifting that thought which is "okay"

however 
1. you can't break existing code.
2. with this new design what is the api contract you are actually trying to build?

assuming most of the implementations don't care about anything but a "pod" then maybe move the createpod deletepod to a common library and have the implementations all call that library?

marcushines (commented):
i don't see changes yet?

cmd/root.go
marcushines: ping don't see changes yet?

topo/node/node.go
```diff
@@ -138,6 +140,20 @@ var (
...
+	pb := n.impl.Proto()
+	log.Infof("Creating Resource for Pod:\n %+v", pb)
+	err := n.impl.CreateNodeResource(ctx, n.kClient, n.namespace)
+	if err == nil {
+		return nil
+		//return fmt.Errorf("failed to create pod resource for %q: %w", pb.Name, err)
+	}
+
+	// For nodes with no resource create support fall back on pod creation
+	return n.CreatePod(ctx)
```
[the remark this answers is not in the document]
marcushines: in general i would prefer to only have "one" api for this but for now i get the intent but i would change this slightly

change your "unimplemented" createNodeResource to return a specific error

status.Errorf(status.Unimplmented, "Unimplemented")

then your check is 
```go
err := n.impl.CreateNodeResource(ctx, n.kClient, n.namespace)
switch status.Code(err) {
case codes.OK:
  return nil
case codes.Unimplemented:
default:
  return err
}
return n.CreatePod(ctx)
```

cmd/root.go
```diff
@@ -113,8 +113,11 @@ func createFn(cmd *cobra.Command, args []string) error {
 	if err := t.Push(cmd.Context()); err != nil {
 		return err
 	}
-	fmt.Fprintf(out, "Topology %q created", topopb.Name)
+	fmt.Fprintf(out, "Topology %q created\n", topopb.Name)
 	r, err := t.Resources(cmd.Context())
+	if err != nil {
+		return fmt.Errorf("%s: %w", cmd.Use, err)
```
marcushines, line 119: just return err

topo/topo.go
```diff
@@ -229,10 +234,12 @@ func (m *Manager) Delete(ctx context.Context) error {
 		n.DeleteService(ctx)
 		// Delete config maps for node
 		n.Delete(ctx)
+		// Delete Resource for node
+		n.DeleteResource(ctx)
 		// Delete Pod
-		if err := m.kClient.CoreV1().Pods(m.tpb.Name).Delete(ctx, n.Name(), metav1.DeleteOptions{}); err != nil {
-			log.Warnf("Error deleting pod %q: %v", n.Name(), err)
-		}
+		//if err := m.kClient.CoreV1().Pods(m.tpb.Name).Delete(ctx, n.Name(), metav1.DeleteOptions{}); err != nil {
```
marcushines, line 240: remove commented code

topo/topo.go
```diff
@@ -210,10 +212,13 @@ func (m *Manager) Push(ctx context.Context) error {
 		if err := n.CreateService(ctx); err != nil {
 			return err
 		}
-		if err := n.CreatePod(ctx); err != nil {
+		if err := n.CreateResource(ctx); err != nil {
 			return err
 		}
-		log.Infof("Node %q created", k)
+		log.Infof("Node %q resource created", k)
+		//if err := n.CreatePod(ctx); err != nil {
```
marcushines, line 219: remove commented code

marcushines (commented):
couple of nits but then you can submit

hashwini-keysight:
Closing this, will send the new one where I incorporated the changes suggested by Marcus

PR #13, opened 2021-07-23
Add support for cli command and initial bringup for generation of self signed certs on arista nodes

marcushines:
This creates the cli command

./kne_cli topology cert ../examples/3node-withtraffic.pb.txt r1 -v=debug

as well as for nodes which provide a self signed cert configuration will try to generate certs on kne bringup
This will add startup time as right now the implementation waits for the pod to become active and then pushes to it


topo/node/ceos/ceos.go
```diff
@@ -34,6 +37,88 @@ func (n *Node) Proto() *topopb.Node {
...
+
+func (n *Node) GenerateSelfSigned(ctx context.Context, ni node.Interface) error {
+	selfSigned := n.pb.GetConfig().GetCert().GetSelfSigned()
+	if selfSigned == nil {
+		log.Infof("%s - no cert config", n.pb.Name)
+		return nil
+	}
+	log.Infof("%s - generating self signed certs", n.pb.Name)
+	log.Infof("%s - Waiting for pod to be running", n.pb.Name)
+	w, err := ni.KubeClient().CoreV1().Pods(ni.Namespace()).Watch(ctx, metav1.ListOptions{
```
alexmasi, line 56: Is there an equivalent to "kubectl wait --for=condition=Ready pods --all -A --timeout=1m" in the go API? Might be easier than using watch and receiving results over the channel.

proto/topo.proto
```diff
@@ -90,6 +92,23 @@ message Config {
...
+    // Also support for CSR and generation workflow.
+  }
+}
+
+message SelfSignedCertCfg {
+  string cert_name = 1;
+  string key_name = 2;
+  uint32 key_size = 3;
+  string common_name = 4;
+  string profile_name = 5;
```
alexmasi, line 109: Some basic comments for the proto fields would be helpful here

topo/node/ceos/ceos.go
```diff
@@ -34,6 +37,88 @@ func (n *Node) Proto() *topopb.Node {
...
+	log.Debugf("%s - enabled - sending %q", n.pb.Name, cmd)
+	if err := g.Send(cmd); err != nil {
+		return err
+	}
+	_, _, err = g.Expect(regexp.MustCompile(`#`), 30*time.Second)
+	if err != nil {
+		return err
+	}
+	cmd = fmt.Sprintf("security pki certificate generate self-signed %s key %s parameters common-name %s\n", selfSigned.CertName, selfSigned.KeyName, n.pb.Name)
+	log.Debugf("key generated - sending %q", cmd)
```
alexmasi, line 110: Nit: Not sure what style we are following for this repo, seems like some log messages begin with capital letters and have punctuation while others don't. Personally I prefer the full sentence style

alexmasi (dismissed):

topo/node/ceos/ceos.go
marcushines: so this is how the command works internally

topo/node/ceos/ceos.go
marcushines: i don't feel debug requires that :)

proto/topo.proto
marcushines: added

shubh90 (approved):

PR #14, opened 2021-07-26
added initial support for nokia srlinux

hellt:
Hi @mhines01 
this PR adds initial support for running Nokia SR Linux on `kne`

1. I wondered if you will be ok to add sysctls configuration for the nodes in the topo (SR linux benefits from certain sysctls to be tuned)?

2. SR Linux system can emulate different HW variants (IXR-6/10, IXR-D1/D2/D3). The decision which HW variant will be emulated is driven by the contents of a file that gets mounted to a POD under a certain path.  
I am curious if you discussed a way to provide node-specific mounts in that context? Can be a config map, for instance.

PS
Note, that [a change](https://github.com/networkop/k8s-topo/pull/15) has been made to `networkop/init-wait` to support srlinux linux interfaces naming. The container image was updated on docker hub, so a `kind load ...` will be enough to make this all work.

The self-generated certs and config push haven't been added yet.


topo/node/srl/srl.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		Labels: map[string]string{
+			"type": topopb.Node_NOKIA_SRL.String(),
+		},
+		Config: &topopb.Config{
+			Image: "srlinux:latest",
+			Command: []string{
+				"/tini",
+			},
+			Args: []string{
```
marcushines, line 89: gah where does that come from?

topo/node/srl/srl.go
```diff
@@ -0,0 +1,111 @@
...
+type Node struct {
+	pb *topopb.Node
+}
+
+func (n *Node) Proto() *topopb.Node {
+	return n.pb
+}
+
+var (
+	spawner = defaultSpawner
```
marcushines, line 35: you don't need this unless you are using expect for anything

hellt: ah, yes, sure, I will add it back when I will have config/tls done
removed in 20baf3b

topo/node/srl/srl.go
hellt: this is a combination of the image's entrypoint

```
docker image inspect --format '{{json .ContainerConfig.Entrypoint}}' srlinux:latest
```

and the command that launches SR Linux process specified in [Software Installation Guide](https://infocenter.nokia.com/public/SRLINUX216R1A/topic/com.srlinux.swinstall/html/swinstall-containers.html?cp=9_2_1#lindapra5isvfaxxqso)

marcushines (approved):

PR #16, opened 2021-07-27
fix service implementation to properly expose node port as well as internal and outside ports

marcushines:
This PR fixes a gap between the knebind in ondatra and the kne service creation.
Services now properly expose inside container / nodeport / outside LB ports 

this makes outside IP and outside port line up with test expectations


proto/topo.proto
```diff
@@ -116,10 +116,14 @@ message SelfSignedCertCfg {
 // provided by the user for which services they would like exposed. Once the
 // service is created KNE will fill in the outside information for the user to
 // access the services.
+// The user should specific inside port for this is the port the container will
```
shubh90, line 119: specify ?

shubh90 (dismissed):

shubh90 (approved):

PR #17, opened 2021-07-28
Add unittests for ceos cert generation

marcushines:


topo/node/ceos/ceos_test.go
```diff
@@ -0,0 +1,267 @@
...
+		wantErr: true,
+	}}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			spawner = tt.spawner
+			mySecond = 0
+			defer func() {
+				spawner = defaultSpawner
+				mySecond = time.Second
+			}()
```
alexmasi, line 254: Nit:

In the case that spawner or mySecond in ceos.go ever change the restoring of the defaults would have to change too. Doing

defSpawner := spawner
defMySecond := mySecond
defer func() {
    spawner = defSpawner
    mySecond = defMySecond
}
spawner = tt.Spawner
mySecond = 0

avoids the potential future code change at the cost of a few additional lines, up to you

topo/node/ceos/ceos_test.go
```diff
@@ -0,0 +1,267 @@
...
+					},
+				},
+			}},
+		}
+		return true, f, nil
+	}
+	ki.PrependWatchReactor("*", reaction)
+	tests := []struct {
+		desc    string
+		wantErr bool
```
alexmasi, line 154: Would prefer this to be a string so errdiff can check the returned error with the substring if we add more test cases to be sure that the function exists for the expected reason

topo/node/ceos/ceos.go
```diff
@@ -45,6 +45,10 @@ func defaultSpawner(command string, timeout time.Duration, opts ...expect.Option
 	return expect.Spawn(command, timeout, opts...)
 }
 
+var (
+	mySecond = time.Second
```
alexmasi, line 49: Nit: Personally I think its more clear when stubbing for testing to call this timeSecond instead of mySecond to be clear this is a stub of the duration time.Second

alexmasi (approved):
LGTM

topo/node/ceos/ceos.go
marcushines: agreed

topo/node/ceos/ceos_test.go
marcushines: done

topo/node/ceos/ceos_test.go
marcushines: ack

alexmasi (approved):

PR #18, opened 2021-08-08
controller managed srlinux resource

hellt:
Hi @marcushines 

Here is a WIP on the srlinux integration with kne via k8s controller.
The controller implementation is in https://github.com/srl-labs/kne-controller

The CRD is generated for the [`Srlinux`](https://github.com/srl-labs/kne-controller/blob/v0.0.0-alpha3/api/v1alpha1/srlinux_types.go#L42) struct which has the [Spec](https://github.com/srl-labs/kne-controller/blob/v0.0.0-alpha3/api/v1alpha1/srlinux_types.go#L24) that consists of the copy of the [NodeConfig](https://github.com/srl-labs/kne-controller/blob/v0.0.0-alpha3/api/v1alpha1/types.go#L3) copied from the `kne` [Node](https://github.com/google/kne/blob/main/proto/topo/topo.pb.go#L176) struct and with added integer field for number of expected interfaces.

Currently, the config that may be passed from the `kne` side is not taken into account, and the defaults are used from the controller package. 

Would appreciate your view on the approach taken here. Since this is my very first stab at k8s api it might be I made quite a few noob mistakes


topo/node/srl/srl.go
```diff
@@ -37,12 +41,65 @@ func (n *Node) ConfigPush(ctx context.Context, ns string, r io.Reader) error {
...
+			Name: n.pb.Name,
+			Labels: map[string]string{
+				"app":  n.pb.Name,
+				"topo": ni.Namespace(),
+			},
+		},
+		Spec: srl.SrlinuxSpec{
+			NumInterfaces: len(ni.Interfaces()),
+			Config: &srl.NodeConfig{
+				Sleep: 0,
```
marcushines, line 62: what does sleep do?

topo/node/srl/srl.go
```diff
@@ -37,12 +41,65 @@ func (n *Node) ConfigPush(ctx context.Context, ns string, r io.Reader) error {
...
+			},
+		},
+	}
+	body, err := json.Marshal(srl)
+	if err != nil {
+		return err
+	}
+
+	err = ni.KubeClient().CoreV1().RESTClient().
```
marcushines, line 71: if you make a clientset to wrap this up you could refer directly to this rather than having to directly use the rest api

see https://github.com/networkop/meshnet-cni/blob/master/api/clientset/v1beta1/topology.go

for an example

topo/node/srl/srl.go
```diff
@@ -37,12 +41,65 @@ func (n *Node) ConfigPush(ctx context.Context, ns string, r io.Reader) error {
...
+	}
+	log.Infof("Created custom resource: %s", n.pb.Name)
+	return nil
 }
 
-func (n *Node) DeleteNodeResource(_ context.Context, _ node.Interface) error {
-	return status.Errorf(codes.Unimplemented, "Unimplemented")
+func (n *Node) DeleteNodeResource(ctx context.Context, ni node.Interface) error {
+	log.Infof("Deleting Srlinux node resource %s", n.pb.Name)
+	err := ni.KubeClient().CoreV1().RESTClient().
```
marcushines, line 89: same here

topo/node/srl/srl.go
hellt: that is the delay in seconds before starting the pod, as per `kne` config

https://github.com/google/kne/blob/570083e95d00fcd4f9ab5e5ceb3a1c95b63c6ee1/proto/topo.proto#L79

topo/node/srl/srl.go
hellt: Will try, sure

hellt: @marcushines I have a question re how to properly build a clientset rest client
for the topology rest client you have a [`New()`](https://github.com/google/kne/blob/570083e95d00fcd4f9ab5e5ceb3a1c95b63c6ee1/topo/topo.go#L63) function that tries to get the rest config from cluster first, and then resorts to `kubecfg`.

Can I do the same from the point in time where you create a custom srlinux resource? Because I don't have access to kubecfg string I think to replicate the same kind of approach

hellt: I added for now the `*rest.Config` to node Interface here b4ec55a

Then I could use the clientset in c3ce4d5

topo/node/srl/srl.go
hellt: implemented with a few additions explained here https://github.com/google/kne/pull/18#discussion_r686024219

topo/node/srl/srl.go
hellt: done

marcushines (approved):

PR #21, opened 2021-08-13
Add deploy to cli

marcushines:
this feature will take a default yaml config file
it will be able to:
deploy a kind cluster
deploy metallb
deploy meshnet


deploy/README.md
```diff
@@ -1,5 +1,29 @@
...
+* Install Metallb
+  * Based on manifests/metallb checked in configurations
+  * Read docker configuration and generate proper configuration file
+  * Wait for deployment to complete
+* Install Meshnet
+  * Based on manifests/meshnet check in configurations
+  * Wait for pods to be healthy
+
+From start to finish this will take ~1 min to bring up the cluster ready for you to deploy a topology to.
+This will stil need images for the devices you wish to use in the topology loaded into the cluster.
```
hellt, line 25: Maybe add more details like:
---

After the deployment is finished additional steps might be required:

* import container images to the cluster image store (unless an image is available for pulling)
* install controllers for nodes which are managed externally

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,458 @@
...
+	if err != nil {
+		return err
+	}
+	deploymentBasePath = filepath.Dir(deploymentBasePath)
+	b, err := ioutil.ReadFile(args[0])
+	if err != nil {
+		return err
+	}
+	dCfg := &DeploymentConfig{}
+	if err := yaml.Unmarshal(b, dCfg); err != nil {
```
hellt, line 414: In yamlv2 there is a strict unmarshalling to catch yaml errs early?

https://pkg.go.dev/gopkg.in/yaml.v2#UnmarshalStrict

It seems in v3 there is no such thing?

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+type ClusterSpec struct {
+	Kind string    `yaml:"kind"`
+	Spec yaml.Node `yaml:"spec"`
+}
+
+type IngressSpec struct {
+	Kind string    `yaml:"kind"`
+	Spec yaml.Node `yaml:"spec"`
```
hellt, line 46: What is the idea behind using yaml.Node and not kindSpec for this field?

deploy/README.md
mhines01: I like to be fairly incremental on the PR's
This is already WAY overly massive of a PR
especially as I add unittests and refactor for that

loading images is also fairly expensive in time so would ideally only load the images needed for a topology so is likely a follow up PR to add that

cmd/deploy/deploy.go
mhines01: i leave this to marshalled based on the kind of the spec
so while this currently supports 
kind cluster + metallb + meshnet

you could write a config and spec implementation for say
Local cluster + calico as ingress + multis 
and it would stay the same form of config

deploy/README.md
hellt: I meant only to provide the above mentioned section. Not the implementation or anything like that

hellt: Meaning that importing images is not needed if the image is available in a repo that is accessible from the cluster (as it will be pulled by the controller)

And that installation of additional controllers might be required after the cluster is installed, given that controllers based operations is envisioned to be the preferred way of launching networking pods

mhines01: updated

cmd/deploy/deploy.go
mhines01: decoder := yaml.NewDecoder(bytes.NewBuffer(b))
	decoder.KnownFields(true)

shubh90 (approved):

deploy/kne/kind.yaml
```diff
@@ -0,0 +1,18 @@
+cluster:
+  kind: Kind
+  spec: 
+    name: kne
+    recycle: True
+    version: 0.11.1
+    image: kindest/node:v1.22.0
```
alexmasi, line 7: I think this needs a "wait" field of 1m. During my testing, the kind cluster creation was much more reliable if it waited (usually less than 20 seconds) to finish setup. Then you would need to override the UnmarshalYAML func

type KindSpec struct {
        Name    string        `yaml:"name"`
        Version string        `yaml:"version"`
        Image   string        `yaml:"image"`
        Retain  bool          `yaml:"retain"`
        Wait    time.Duration `yaml:"wait"`
        Kubecfg string        `yaml:"kubecfg"`
}

func (k *KindSpec) UnmarshalYAML(unmarshal func(interface{}) error) error {
        type alias struct {
                Name    string `yaml:"name"`
                Version string `yaml:"version"`
                Image   string `yaml:"image"`
                Retain  bool   `yaml:"retain"`
                Wait    string `yaml:"wait"`
                Kubecfg string `yaml:"kubecfg"`
        }

        var a alias
        if err := unmarshal(&a); err != nil {
                return err
        }
        t, err := time.ParseDuration(a.Wait)
        if err != nil {
                return err
        }
        k.Name = a.Name
        k.Version = a.Version
        k.Image = a.Image
        k.Retain = a.Retain
        k.Wait = t
        k.Kubecfg = a.Kubecfg
        return nil
}

cmd/deploy/deploy_test.go
```diff
@@ -0,0 +1,125 @@
+package deploy
```
alexmasi, line 1: I misunderstood what you said yesterday about adding contributing to this branch, here is what I threw together yesterday: https://paste.googleplex.com/6488835253862400. Never uploaded the commit though :(

If there are any sections you prefer feel free to add them. The only major thing was the kind waiting I added a comment for already

cmd/deploy/deploy.go
```diff
@@ -0,0 +1,469 @@
...
+	if err != nil {
+		return err
+	}
+	kClient, err := kubernetes.NewForConfig(rCfg)
+	if err != nil {
+		return err
+	}
+	d.Ingress.SetKClient(kClient)
+	log.Infof("Validating cluster health")
+	if err := d.Ingress.Deploy(cmd.Context()); err != nil {
```
alexmasi, line 450: In my experience it takes around 4 attempts to successfully deploy.

Maybe we should add the retry logic here

deploy/kne/kind.yaml
```diff
@@ -0,0 +1,18 @@
...
+ingress:
+  kind: MetalLB
+  spec:
+    manifests: ../../manifests/metallb
```
alexmasi, line 11: Does this work if you do not call the CLI from the dir its built in? I was having some trouble figuring out how best to resolve paths like this

cmd/deploy/deploy.go
mhines01: hmm i have had zero failures so far with this setup

i don't want to guess at retrying things until we have data

deploy/kne/kind.yaml
mhines01: see the code i look at the path that the yaml loaded from then use that directory as relative path from there

deploy/kne/kind.yaml
mhines01: you shouldn't wait you should watch

right now i am just using the fact i get a client as the validation but we can decide what watches should be there like we could check the control plane pod status but so far i don't have an example of it not working

shubh90 (approved):

PR #23, opened 2021-08-17
Add link for cPTX from github.

kmo7:
* This allows cPTX topo deployment from KNE's github repo.

* Added copyright string for cPTX per Juniper Legal team request.

### DIFF
--- a/topo/node/cptx/cptx.go
+++ b/topo/node/cptx/cptx.go
@@ -1,4 +1,5 @@
 // Juniper cPTX for KNE
+// Copyright (c) Juniper Networks, Inc., 2021. All rights reserved.
 
 package cptx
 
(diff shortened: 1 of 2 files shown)


kmo7:
Hi Marcus, please review this at your convenience. Thanks, Kaveh

marcushines:
just FYI - had to ask internally if there was any reason we could care about you adding a juniper copyright into this file.
I think the answer is we don't care but just wanted clarification before merging

kmo7:
Ack. Marcus, guessed that was the reason for waiting to merge. This was a request from our legal team, which as you know has also approved the google CLA.
Please let me know when your legal team decides.  Thanks, Kaveh



Juniper Business Use Only

marcushines (approved):
Checked with legal and adding a copyright here is fine

PR #29, opened 2021-08-19
Add reset config to nodes which support and add cli command

mhines01:
This adds 

kne_cli topology reset <config> <device>

to reset the configuration of the node to a default config

if no devices are provides all nodes in the topology will be attempted
--skip will skip nodes which don't support resetcfg but will continue through all nodes

### DIFF
--- a/cmd/topology/topology.go
+++ b/cmd/topology/topology.go
@@ -72,8 +72,8 @@ var (
 )
 
 func resetCfgFn(cmd *cobra.Command, args []string) error {
-	if len(args) != 1 {
-		return fmt.Errorf("%s: missing args", cmd.Use)
+	if len(args) < 1 || len(args) > 2 {
+		return fmt.Errorf("%s: invalid args", cmd.Use)
 	}
 	topopb, err := topo.Load(args[0])
 	if err != nil {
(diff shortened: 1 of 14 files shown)


cmd/topology/topology.go
```diff
@@ -44,8 +60,64 @@ var (
...
+	}
+)
+
+var (
+	skipReset bool
 )
 
+func resetCfgFn(cmd *cobra.Command, args []string) error {
+	if len(args) != 1 {
+		return fmt.Errorf("%s: missing args", cmd.Use)
```
alexmasi, line 76: Won't this cause failure if a user specifies a specific device as well? I think you meant

if len(args) < 1

cmd/topology/topology.go
```diff
@@ -44,8 +60,64 @@ var (
...
+	if err != nil {
+		return err
+	}
+	t, err := topo.New(s, topopb)
+	if err != nil {
+		return fmt.Errorf("%s: %w", cmd.Use, err)
+	}
+	ctx := cmd.Context()
+	t.Load(ctx)
+	var nodes []*node.Node
```
alexmasi, line 92: I think this should simplify this block a bit:

nodes := t.Nodes()
if len(args) > 1 {
    n, err := t.Node(args[1])
    if err != nil {
        return err
    }
    nodes = []*node.Node{n}
}

cmd/topology/topology.go
```diff
@@ -44,8 +60,64 @@ var (
...
+	if len(args) == 1 {
+		nodes = t.Nodes()
+	} else {
+		n, err := t.Node(args[1])
+		if err != nil {
+			return err
+		}
+		nodes = append(nodes, n)
+	}
+	var resetable []node.Reseter
```
alexmasi, line 102: nit: resettable and later in the interface its resetter

alexmasi (approved):

cmd/topology/topology.go
mhines01: fixed

alexmasi (approved):

cmd/topology/topology.go
mhines01: done

cmd/topology/topology.go
```diff
@@ -44,8 +60,62 @@ var (
...
+	ctx := cmd.Context()
+	t.Load(ctx)
+	nodes := t.Nodes()
+	if len(args) > 1 {
+		n, err := t.Node(args[1])
+		if err != nil {
+			return err
+		}
+		nodes = append(nodes, n)
+	}
```
alexmasi, line 99: This will add the specified node to the entire collection of nodes a second time.

marcushines: grr this why i didn't want to have that pattern

shubh90 (approved):

PR #30, opened 2021-08-20
Add Version/model/vendor into proto.

mhines01:
This PR is the first step in breaking up the type field into a triple of 

Vendor/Model/Version

this will allow vendors to provide different implemenation details for each model (hardware emulation) or os type


proto/topo.proto
```diff
@@ -29,6 +29,19 @@ message Topology {
   repeated Link links = 3; // connections between Nodes.
 }
 
+// Vendor of the node. Topology manager uses this enum dispatch the node to
```
alexmasi, line 32: topo manager uses this enum _to_ dispatch

alexmasi (dismissed):

marcushines (approved):

PR #33, opened 2021-08-24
add Unittests for reset

marcushines:


cmd/topology/topology.go
```diff
@@ -149,12 +154,24 @@ func resetCfgFn(cmd *cobra.Command, args []string) error {
...
+		switch v := cd.(type) {
+		case *tpb.Config_Data:
+			b = v.Data
+		case *tpb.Config_File:
+			var cPath string
+			if strings.HasPrefix(v.File, "/") {
+				cPath = v.File
+			} else {
+				cPath = filepath.Join(bp, v.File)
+			}
```
alexmasi, line 165: cPath := v.File
			if !filepath.IsAbs(cPath) {
				cPath = filepath.Join(bp, v.File)
			}

cmd/topology/topology_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/h-fam/errdiff"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/prototext"
+	kfake "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/rest"
+)
+
+func NewNR(pb *topopb.Node) (node.Implementation, error) {
+	return &notResetable{pb: pb}, nil
```
alexmasi, line 27: nit: resettable here and elsewhere

cmd/topology/topology_test.go
```diff
@@ -0,0 +1,248 @@
...
+	return &resetable{&notResetable{pb: pb}, nil}, nil
+}
+
+func writeTopology(t *testing.T, topo *topopb.Topology) (*os.File, func()) {
+	t.Helper()
+	f, err := ioutil.TempFile("", "reset")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(t.Name())
```
alexmasi, line 77: Is this line needed? Not sure what exactly is being removed, the test name? The file cleanup func is returned so not sure

cmd/topology/topology_test.go
```diff
@@ -0,0 +1,248 @@
...
+					File: "dne",
+				},
+			},
+		}, {
+			Name: "notresettable1",
+			Type: topopb.Node_Type(1002),
+		}},
+	}
+	fConfigDNE, closer := writeTopology(t, tWithConfigDNE)
+	defer closer()
```
alexmasi, line 180: Do these defers properly clean all the files up? Not sure if there are any issues defer a func and then overwriting the func and deferring it again. Probably not but want to make sure this is working as intended

cmd/topology/topology_test.go
```diff
@@ -0,0 +1,248 @@
...
+		desc: "valid topology push with relative file location",
+		args: []string{"reset", fConfigRelative.Name(), "--skip", "--push"},
+	}, {
+		desc:    "valid topology push with config DNE",
+		args:    []string{"reset", fConfigDNE.Name(), "--skip", "--push"},
+		wantErr: "no such file or directory",
+	}}
+
+	rCmd := New()
+	var s string
```
alexmasi, line 226: Where is this being used?

cmd/topology/topology_test.go
```diff
@@ -0,0 +1,248 @@
...
+		topo.WithClusterConfig(&rest.Config{}),
+		topo.WithKubeClient(kfake.NewSimpleClientset()),
+		topo.WithTopoClient(tf),
+	}
+	defer func() {
+		opts = origOpts
+	}()
+	rCmd.PersistentFlags().StringVar(&s, "kubecfg", "", "")
+	buf := bytes.NewBuffer([]byte{})
+	rCmd.SetOut(buf)
```
alexmasi, line 242: Is this necessary to capture the output if there is no validation being done? Maybe this is required to prevent out from being sent to stdout just not sure

topo/topo.go
```diff
@@ -61,37 +61,65 @@ type Manager struct {
...
+			}
 		}
+		m.rCfg = rCfg
 	}
-	// create the clientset
-	kClient, err := kubernetes.NewForConfig(rCfg)
-	if err != nil {
-		return nil, err
+	if m.kClient == nil {
+		kClient, err := kubernetes.NewForConfig(m.rCfg)
```
alexmasi, line 109: Adding a log for creating default kclient or something similar would be helpful here (similar to the case when rCfg is nil)

topo/topo.go
```diff
@@ -61,37 +61,65 @@ type Manager struct {
...
+		if err != nil {
+			return nil, err
+		}
+		m.kClient = kClient
 	}
-	tClient, err := topologyclientv1.NewForConfig(rCfg)
-	if err != nil {
-		return nil, err
+	if m.tClient == nil {
+		tClient, err := topologyclientv1.NewForConfig(m.rCfg)
```
alexmasi, line 116: Same as above

alexmasi (dismissed):

cmd/topology/topology_test.go
marcushines: no idea where that came frome...

cmd/topology/topology_test.go
marcushines: i am not overwriting anything just reuse a pointer and call it - the fact that the address changes doesn't change anything

cmd/topology/topology_test.go
marcushines: yeah i just makes the test log no noisy

topo/topo.go
marcushines: i don't see that being super useful it is just a client - we don't generally "log" creating grpc stub and such

alexmasi: I was thinking it could be useful if the user tries to provide one of the newly added options and the log would show that a default value is being used if the intended option wasn't provided.

But I see your point, the in cluster kubecfg logs provide more insight than simply a new kclient was created. SGTM

cmd/topology/topology_test.go
marcushines: hmm not sure why lint didn't find that but good catch

cmd/topology/topology_test.go
marcushines: done

cmd/topology/topology_test.go
marcushines: it was actgually used for holding the persistent flag but found a better method

cmd/topology/topology.go
marcushines: done

alexmasi (approved):

PR #34, opened 2021-08-27
draft fixup of scrapli failed/error setup

carlmontanari:
- no more bleeding in of the test helper (created an interface for system transport and then just let the test transport implement that so we are always good with test and with "real" transport)
- `Failed` is an error now rather than a bool so we can just return that directly
- currently pinned to branch im working on doing some cleanup on so just a draft for now, but if this looks good will update pin once that branch gets merged

@hellt @marcushines

### DIFF
--- a/topo/node/ceos/ceos.go
+++ b/topo/node/ceos/ceos.go
@@ -21,8 +21,6 @@ import (
 	"io/ioutil"
 	"time"
 
-	scraplitest "github.com/scrapli/scrapligo/util/testhelper"
-
 	topopb "github.com/google/kne/proto/topo"
 	"github.com/google/kne/topo/node"
 	scraplibase "github.com/scrapli/scrapligo/driver/base"
@@ -40,9 +38,6 @@ import (
 
 // ErrIncompatibleCliConn raised when an invalid scrapligo cli transport type is found.
 var ErrIncompatibleCliConn = errors.New("incompatible cli connection in use")
-
-// ErrCliOperationFailed raised when scrapligo reports failure in some cli operation.
-var ErrCliOperationFailed = errors.New("cli operation failed")
 
 func New(pb *topopb.Node) (node.Implementation, error) {
 	cfg := defaults(pb)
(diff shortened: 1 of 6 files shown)


topo/node/ceos/ceos.go
```diff
@@ -81,17 +76,14 @@ func (n *Node) WaitCLIReady() error {
...
 func (n *Node) PatchCLIConnOpen(ns string) error {
-	switch t := n.cliConn.Transport.(type) {
-	case *scraplitest.TestingTransport:
-		t.ExecCmd = "kubectl"
-		t.OpenCmd = []string{"exec", "-it", "-n", ns, n.pb.Name, "--", "Cli"}
-	case *scraplitransport.System:
-		t.ExecCmd = "kubectl"
-		t.OpenCmd = []string{"exec", "-it", "-n", ns, n.pb.Name, "--", "Cli"}
-	default:
+	t, ok := interface{}(n.cliConn.Transport.Impl).(scraplitransport.SystemTransport)
```
hellt, line 79: can it be
```go
t, ok := n.cliConn.Transport.Impl.(scraplitransport.SystemTransport)
```
?

carlmontanari: Good call... was wrapped around the axle there for a minute -- fixed 😁

carlmontanari:
Going to make this as "ready for review" but with the caveat that if this all looks good I will merge that branch in scrapligo and update this pr w/ a pin to that main branch or a release if I finally get my life right and make one 😁

topo/node/ceos/ceos.go
```diff
@@ -171,18 +163,14 @@ func (n *Node) GenerateSelfSigned(ctx context.Context, ni node.Interface) error
...
+	resp, err := n.cliConn.SendConfigs(cfgs)
 	if err != nil {
 		return err
 	}
 
-	if r.Failed() {
-		return ErrCliOperationFailed
-	}
-
 	log.Infof("%s - finshed cert generation", n.pb.Name)
```
alexmasi, line 171: Do you want this log to happen even if resp.Failed? Previously this message would only log if the resp was not marked failed

carlmontanari: ah yes -- I dont think we should log if it failed, will push an update shortly!

alexmasi:
Okay this LGTM, let me know when you merge the scrapligo changes and I'll approve this

carlmontanari:
@alexmasi awesome -- just merged that branch and released v0.1.0 and updated this to pin to that!

alexmasi (approved):

topo/node/ceos/ceos.go
```diff
@@ -171,18 +163,16 @@ func (n *Node) GenerateSelfSigned(ctx context.Context, ni node.Interface) error
 		),
 	}
 
-	r, err := n.cliConn.SendConfigs(cfgs)
+	resp, err := n.cliConn.SendConfigs(cfgs)
```
marcushines, line 166: still haven't removed the "multiple error" pattern i thought that was a goal of this pr?

carlmontanari: the goals as I understood it were to fix the transport interfaces up so that the testing stuff didn't leak into production bits as well as to make `Failed` return an error (or nil ofc) instead of a bool. that `Failed` error also now contains the contextual info about the "error" that was encountered.

the long standing scrapli pattern here is that we only raise exception (in python ofc) or return an error if the connection encounters an error, so unless that happens we always return a response. if we get that response we have that extra feature to test if the remote device produced some application error (ie invalid command/wrong port/whatever). so basically the api contract w/in scrapli is to *not* error out if the inputs were delivered properly and outputs read off the wire properly.

we can remove the extra "level of checking" here if you prefer but this pattern has been working quite well for scrapli so the idea was brought into go version and felt useful/valuable here too.

PR #35, opened 2021-08-31
Add a kubectl install check before running kne_cli deploy deploy command

shubh90:


cmd/deploy/deploy.go
```diff
@@ -116,6 +117,9 @@ func deployFn(cmd *cobra.Command, args []string) error {
 	if len(args) != 1 {
 		return fmt.Errorf("%s: missing args", cmd.Use)
 	}
+	if _, err := os.Stat("/usr/local/bin/kubectl"); err != nil {
```
alexmasi, line 120: Use lookpath, kubectl could be installed at a different filepath: https://pkg.go.dev/os/exec#example-LookPath

shubh90: Done

alexmasi (approved):

PR #54, opened 2021-09-30
Update README.md

marcushines:


shubh90 (approved):

hellt:
You have put go1.17 in readme, yet go.mod is set to 0.16. 
Does it mean we will transition to 0.17?

PR #59, opened 2021-10-07
Propagate node resource deployment status

anjan-keysight:
Enhancement for propagating node deployment status to cli.

### DIFF
--- a/topo/node/srl/srl.go
+++ b/topo/node/srl/srl.go
@@ -95,6 +95,10 @@ func (n *Node) CreateNodeResource(ctx context.Context, ni node.Interface) error 
 	return nil
 }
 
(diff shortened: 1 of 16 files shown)


topo/node/srl/srl.go
```diff
@@ -95,6 +95,10 @@ func (n *Node) CreateNodeResource(ctx context.Context, ni node.Interface) error
 	return nil
 }
 
+func (n *Node) GetNodeResourceStatus(_ context.Context, _ node.Interface) (node.NodeStatus, error) {
+	return node.NodeStatus{}, status.Errorf(codes.Unimplemented, "Unimplemented")
```
hellt, line 99: I wonder should this be `nil` to remove allocations in the non implemented case?

anjan-keysight: Yes we can make it a pointer and set it to nil. I will update.

PR #61, opened 2021-10-07
Error propagation with review comment incorporated

anjan-keysight:
Error propagation enhancements. Review comment incorporated.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -162,6 +162,9 @@ func createFn(cmd *cobra.Command, args []string) error {
 	if err := t.Push(cmd.Context()); err != nil {
 		return err
 	}
+	if err := t.CheckNodeStatus(cmd.Context()); err != nil {
+		return err
+	}
 	fmt.Fprintf(out, "Topology %q created\n", topopb.Name)
 	r, err := t.Resources(cmd.Context())
 	if err != nil {
--- a/topo/node/ixia/ixia.go
+++ b/topo/node/ixia/ixia.go
@@ -12,7 +12,7 @@ import (
 )
 
 type IxiaSpec struct {
-	Config string `json:"config,omitempty"`
+	Config  string `json:"config,omitempty"`
 	Version string `json:"version,omitempty"`
 }
 
@@ -20,7 +20,9 @@ type Ixia struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec IxiaSpec `json:"spec,omitempty"`
+	Spec   IxiaSpec        `json:"spec,omitempty"`
+	// This is a temporary fix until Ixia operator is made public and IxiaTG type can be referenced
+	Status node.NodeStatus `json:"status,omitempty"`
 }
 
 func New(pb *topopb.Node) (node.Implementation, error) {
@@ -79,6 +81,31 @@ func (n *Node) CreateNodeResource(ctx context.Context, ni node.Interface) error 
 	}
 	log.Infof("Created custom resource: %s", n.pb.Name)
 	return nil
+}
+
+func (n *Node) GetNodeResourceStatus(ctx context.Context, ni node.Interface) (node.NodeStatus, error) {
+	ixiaNode := Ixia{}
+	err := error(nil)
+	res := ni.KubeClient().CoreV1().RESTClient().
+		Get().
+		AbsPath("/apis/network.keysight.com/v1alpha1").
+		Namespace(ni.Namespace()).
+		Resource("Ixiatgs").
+		Name(n.pb.Name).
+		Do(ctx)
+
+	if err = res.Error(); err == nil {
+		if resraw, err := res.Raw(); err == nil {
+			err = json.Unmarshal(resraw, &ixiaNode)
+		}
+	}
+
+	if err != nil {
+		log.Error(err)
+		return node.NodeStatus{}, err
+	}
+
+	return ixiaNode.Status, nil
 }
 
 func (n *Node) DeleteNodeResource(ctx context.Context, ni node.Interface) error {
--- a/topo/node/ixia/ixia.go
+++ b/topo/node/ixia/ixia.go
@@ -83,7 +83,7 @@ func (n *Node) CreateNodeResource(ctx context.Context, ni node.Interface) error 
 	return nil
 }
 
-func (n *Node) GetNodeResourceStatus(ctx context.Context, ni node.Interface) (node.NodeStatus, error) {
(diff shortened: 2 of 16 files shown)


anjan-keysight:
@marcushines enhancements for node deployment error propagation.

cmd/root.go
```diff
@@ -162,6 +162,9 @@ func createFn(cmd *cobra.Command, args []string) error {
 	if err := t.Push(cmd.Context()); err != nil {
 		return err
 	}
+	if err := t.CheckNodeStatus(cmd.Context()); err != nil {
```
marcushines, line 165: so I am not sure we "always" want to wait here i would feel this likely should be a flag

topo/node/ixia/ixia.go
```diff
@@ -81,6 +83,31 @@ func (n *Node) CreateNodeResource(ctx context.Context, ni node.Interface) error
 	return nil
 }
 
+func (n *Node) GetNodeResourceStatus(ctx context.Context, ni node.Interface) (*node.NodeStatus, error) {
```
marcushines, line 86: This should probalbly just be 

Status(ctx)

node now provides the node.Interface so you don't need to pass it in anymore

topo/node/node.go
```diff
@@ -38,6 +38,9 @@ type Implementation interface {
 	// CreateNodeResource provides a custom implementation of pod creation
 	// for a node type. Requires context, Kubernetes client interface and namespace.
 	CreateNodeResource(context.Context, Interface) error
+	// GetNodeResource provides a custom implementation of accessing vendor node status.
+	// Requires context, Kubernetes client interface and namespace.
+	GetNodeResourceStatus(context.Context, Interface) (*NodeStatus, error)
```
marcushines, line 43: just Status

topo/node/node.go
```diff
@@ -72,6 +75,11 @@ func Register(t topopb.Node_Type, fn NewNodeFn) {
 	mu.Unlock()
 }
 
+type NodeStatus struct {
```
marcushines, line 78: type Status since it will read as node.Status in other contexts

topo/topo.go
```diff
@@ -256,6 +258,55 @@ func (m *Manager) Push(ctx context.Context) error {
...
+func (m *Manager) CheckNodeStatus(ctx context.Context) error {
+	foundAll := false
+	failed := false
+	nodes := make(map[string]*node.Node)
+	for k, n := range m.nodes {
+		nodes[k] = n
+	}
+
+	// Check until end state or 3 sec expired
+	for i := 0; i < 30 && !foundAll; i++ {
```
marcushines, line 271: Why are we saying wait for 3 secs? feels like we should either block until a state or have a flag provided timeout to wait?

what is our guarantee that this everything is in the right status?

topo/node/ixia/ixia.go
anjan-keysight: Will update.

topo/node/node.go
anjan-keysight: Sure

topo/node/node.go
anjan-keysight: Will update.

topo/topo.go
anjan-keysight: Will provide a timeout <seconds> flag.

For successful status the Reason would be empty. Will leave it up to node specific implementation, to interpret overall successful status.

anjan-keysight: Just to clarify, for Ixia we will verify Pod and Service status before we do a status update Success.

cmd/root.go
```diff
@@ -80,7 +82,7 @@ func init() {
 
 var (
 	createCmd = &cobra.Command{
-		Use:       "create <topology file>",
+		Use:       "create <topology file> [timeout sec]",
```
alexmasi, line 85: timeout is a flag here, probably best to remove this part of the comment and rely on the auto documentation by cobra

topo/node/ixia/ixia.go
```diff
@@ -83,6 +92,39 @@ func (n *Node) Create(ctx context.Context) error {
 	return nil
 }
 
+func (n *Node) Status(ctx context.Context) (corev1.PodPhase, error) {
+	ixiaNode := Ixia{}
+	var phase corev1.PodPhase
+	err := error(nil)
+	pod, _ := n.KubeClient.CoreV1().Pods(n.Namespace).Get(ctx, n.Name(), metav1.GetOptions{})
```
alexmasi, line 99: Is an error return value being ignored here? I'm not sure the signature but the error should be handled in some way and not be ignored

topo/node/ixia/ixia.go
```diff
@@ -83,6 +92,39 @@ func (n *Node) Create(ctx context.Context) error {
...
+	if pod != nil && pod.Status.Phase == "Running" {
+		return pod.Status.Phase, nil
+	}
+
+	if pod != nil {
+		phase = pod.Status.Phase
+	} else {
```
alexmasi, line 106: A simpler way to write this is

phase := corev1.PodUnknown
if pod != nil {
    phase = pod.Status.Phase
}

And then you can move the if running check after this.

if pod.Status.Phase == "Running" { ... }

Is there a constant or enum for the "Running" status. Its better than using a raw string here

topo/node/ixia/ixia.go
```diff
@@ -83,6 +92,39 @@ func (n *Node) Create(ctx context.Context) error {
 	return nil
 }
 
+func (n *Node) Status(ctx context.Context) (corev1.PodPhase, error) {
+	ixiaNode := Ixia{}
+	var phase corev1.PodPhase
+	err := error(nil)
```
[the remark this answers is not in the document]
anjan-keysight: Incorporated changes

topo/node/ixia/ixia.go
anjan-keysight: Handled error case.

anjan-keysight:
@alexmasi I have incorporated the review comments.

topo/node/ixia/ixia.go
```diff
@@ -83,6 +93,42 @@ func (n *Node) Create(ctx context.Context) error {
...
+	var phase corev1.PodPhase
+	var err error
+	phase = corev1.PodUnknown
+	pod, err := n.KubeClient.CoreV1().Pods(n.Namespace).Get(ctx, n.Name(), metav1.GetOptions{})
+	if err != nil {
+		// Ignore not found error, because operator may not have picked up immediately
+		if !errapi.IsNotFound(err) {
+			return phase, err
+		}
+	} else if pod != nil {
```
marcushines, line 107: no need for else

marcushines (approved):

PR #65, opened 2021-10-12
Update cloudbuild configuration making use of added `deploy` support for pulling images from GAR

alexmasi:
Changes:

- update cloudbuild to pull desired version of gh repo
- use ubuntu instead of debian
- fork manifests and deployment yaml for internal and external versions (changes where images are pulled from)
- add private google artifact registry support for created kind clusters
- reworked execer to expose stdout/err destinations, added more robust fake
- unit tests

TODO in future change:
- add option to use a different init container to support internal use (add option in topo proto?)


cmd/deploy/specs.go
```diff
@@ -130,10 +150,85 @@ func (k *KindSpec) Deploy(ctx context.Context) error {
...
+		}
+		if err := k.execer.exec("docker", "exec", node, "systemctl", "restart", "kubelet.service"); err != nil {
+			return err
+		}
+	}
+	log.Infof("Setup credentials for accessing GAR locations %v in kind cluster", k.GoogleArtifactRegistries)
+	return nil
+}
+
+func writeDockerConfig(path string, registries []string) error {
```
marcushines, line 214: would a template be nicer to use here?

manifests/meshnet/external/external/base/crd.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apiextensions.k8s.io/v1
```
marcushines, line 1: err wrong path?

alexmasi: Fixed.

cmd/deploy/specs.go
```diff
@@ -59,7 +75,8 @@ type KindSpec struct {
 	Wait             time.Duration `yaml:"wait"`
 	Kubecfg          string        `yaml:"kubecfg"`
 	DeployWithClient bool          `yaml:"deployWithClient"`
-	execer           func(string, ...string) error
+	GoogleArtifactRegistries     []string      `yaml:"googleArtifactRegistries"`
+	execer           execerInterface
```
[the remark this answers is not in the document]
alexmasi: I would argue this is no less "pure" than before, adding the extra layer is an option that I can implement. If there is no reason for an execer to belong to each spec (and if each spec may need access to it) then what about a global variable:

var (
    execer execerInterface = exec.NewExecer()
)

Specs in the file can directly call execer.Exec() and tests can stub execer to equal exec.NewFakeExecer()

That would work in my head but not sure which is better, any thoughts?

cmd/deploy/exec.go
```diff
@@ -0,0 +1,61 @@
+package exec
```
marcushines, line 1: go build will not like this

cmd/deploy/specs.go
```diff
@@ -130,20 +157,83 @@ func (k *KindSpec) Deploy(ctx context.Context) error {
...
+	if len(k.GoogleArtifactRegistries) == 0 {
+		log.Debug("No registries require setup")
+		return nil
+	}
+	tempDockerDir, err := os.MkdirTemp("", "kne_kind_docker")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tempDockerDir)
+	originalConfig := os.Getenv(dockerConfigEnvVar)
```
marcushines, line 177: i would maybe suggest a docstring here on why and what you are doing here - it probably isn't super obvious why this is necessary

alexmasi: done

cmd/deploy/exec.go
alexmasi: done

cmd/deploy/specs.go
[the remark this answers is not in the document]
alexmasi: done

PR #77, opened 2021-11-12
Add large timeout for config commit - junos

nitinsoniism:
If a config to be pushed is incredibly large (a few thousand lines),
commit operation may take unusual amount of time. Setting a
large timeout to avoid channel timeout.

### DIFF
--- a/topo/node/cptx/cptx.go
+++ b/topo/node/cptx/cptx.go
@@ -96,6 +96,8 @@ func (n *Node) SpawnCLIConn(ns string) error {
 		scraplibase.WithAuthBypass(true),
 		// disable transport timeout
 		scraplibase.WithTimeoutTransport(0),
+		// for committing a very large config
+		scraplibase.WithTimeoutOps(300 * time.Second),
 	)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


topo/node/cptx/cptx.go
```diff
@@ -96,6 +96,8 @@ func (n *Node) SpawnCLIConn(ns string) error {
 		scraplibase.WithAuthBypass(true),
 		// disable transport timeout
 		scraplibase.WithTimeoutTransport(0),
+		// for committing a very large config
+		scraplibase.WithTimeoutOps(300 * time.Second),
```
alexmasi, line 100: Style nit:

I think it would be better to define a var block at the top of the file with the different timeout durations used in this file.

i.e.

var (
    waitForCLITimeout = 500 * time.Second
    scrapliDriverTimeout = 300 * time.Second
)

and then referenced as needed. IMO it would add clarity

alexmasi (approved):

topo/node/cptx/cptx.go
nitinsoniism: Agree (y). I have taken care of it in the next commit.

alexmasi (approved):

PR #79, opened 2021-11-28
Sm controller

shubh90:


proto/controller.proto
```diff
@@ -0,0 +1,33 @@
...
+option go_package = "github.com/google/kne/proto/controller";
+
+// Topology manager service definition.
+service TopologyManager {
+  rpc CreateTopology (CreateTopologyRequest) returns(CreateTopologyResponse) {}
+}
+
+// Request message to create a topology from a file.
+message CreateTopologyRequest {
+  string file_name = 1;
```
marcushines, line 27: how is that going to work... the cli will have the file locally how are you getting the file onto the service running in k8s?

proto/controller.proto
```diff
@@ -0,0 +1,33 @@
...
+}
+  
+// Returns create topology response.
+message CreateTopologyResponse {
+  string message = 1;
```
marcushines, line 32: what is a message?

proto/controller.proto
shubh90: Changed CreateTopologyRequest to take in topology proto

proto/controller.proto
shubh90: doing empty response right now. But, was also considering CreateTopologyResponse to send topology proto back.

proto/controller.proto
```diff
@@ -22,13 +22,47 @@ option go_package = "github.com/google/kne/proto/controller";
...
+  rpc DeleteTopology (DeleteTopologyRequest) returns(DeleteTopologyResponse) {}
+  rpc ShowTopology (ShowTopologyRequest) returns(ShowTopologyResponse) {}
 }
 
-// Request message to create a topology from a file.
+enum NodeState {
+  STATE_UNKNOWN = 0;
+  STATE_FAILED = 1;
+  STATE_RUNNING = 2;
+  STATE_DELETED = 3;
```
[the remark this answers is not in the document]
shubh90: Updated states to: CREATING -> CONTOLLER_AWAIT -> STARTING -> RUNNING -> ERROR

proto/controller.proto
```diff
@@ -0,0 +1,34 @@
...
+  rpc CreateTopology (CreateTopologyRequest) returns(CreateTopologyResponse) {}
+}
+
+// Request message to create a topology from a file.
+message CreateTopologyRequest {
+  topo.Topology topology = 1;
+}
+  
+// Returns create topology response.
+message CreateTopologyResponse {
```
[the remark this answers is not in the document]
shubh90: We return the state and the name of the topology created

marcushines (approved):

PR #84, opened 2022-01-12
Update ceos node with new pattern

shubh90:


topo/node/ceos/ceos.go
```diff
@@ -263,38 +269,54 @@ func defaults(pb *tpb.Node) *tpb.Node {
...
-		Labels: map[string]string{
+		}
+	}
+	if pb.Labels == nil {
+		pb.Labels = map[string]string{
 			"type":    tpb.Node_ARISTA_CEOS.String(),
 			"vendor":  tpb.Vendor_ARISTA.String(),
 			"model":   pb.Model,
 			"os":      pb.Os,
 			"version": pb.Version,
```
alexmasi, line 280: Should the map values here and in "Env" only get merged if the initial field is nil or should these values be added to the map regardless of the value?

marcushines: they should be merged

marcushines: if they aren't "present" overwrite them

shubh90: Done. Made "Label" and "Env" merge and overwrite with defaults if not present

marcushines (approved):

PR #89, opened 2022-01-20
add node type for cisco 8000

mojiiba:
This pull request adds support for dockerized emulation of cisco 8000 product family (e8000). The main changes are:

- add node type E8000 to represent e8000 products. I understand that node type will be deprecated in the future. However, we still need the node type to be added, since Ondatra KNE binding does not work without this (will send another pull for Ondatra). We do not use node type for node registration, so we can remove the added node types if Ondatra does not require that. 
- Add interface mapping for different models of E8000  
- Add unit tests to test the new changes
- Replace the xrd package with Cisco packages that support both node type xrd and e8000.
- Add sample kne cluster using e8000 and xrd nodes


mojiiba:
Please note that I have already signed the CLA.

mojiiba:
Will fix the cla issue and reopn ....

PR #90, opened 2022-01-20
Update controller service proto with creating cluster design modifica…

shubh90:
…tions


proto/controller.proto
```diff
@@ -24,14 +24,87 @@ service TopologyManager {
...
+}
+
+// Request message to delete a cluster.
+message DeleteClusterRequest {
+  string name = 1;
+}
+
+// Returns delete cluster response.
+message DeleteClusterResponse {
+  string response = 1;
```
alexmasi, line 89: what is this field for? same with the deletetoporesponse message

proto/controller.proto
```diff
@@ -24,14 +24,87 @@ service TopologyManager {
   rpc CreateTopology (CreateTopologyRequest) returns(CreateTopologyResponse) {}
   rpc DeleteTopology (DeleteTopologyRequest) returns(DeleteTopologyResponse) {}
   rpc ShowTopology (ShowTopologyRequest) returns(ShowTopologyResponse) {}
+  rpc CreateCluster (CreateClusterRequest) returns(CreateClusterResponse) {}
```
alexmasi, line 27: All of these RPCs should have a doc string one liner

proto/controller.proto
```diff
@@ -42,7 +115,7 @@ message CreateTopologyRequest {
 // Returns create topology response.
 message CreateTopologyResponse {
   string topology_name = 1;
-  NodeState state = 2;
+  TopologyState state = 2;
```
alexmasi, line 118: we may have to add another field here for the kubecfg (so the topo knows which cluster to create in)

proto/controller.proto
shubh90: Done

proto/controller.proto
shubh90: Made empty response

proto/controller.proto
shubh90: Done

alexmasi (approved):

PR #91, opened 2022-01-21
modify cisco nodes

mojiiba:
This pull request adds support for dockerized emulation of cisco 8000 product family (e8000). The main changes are:

- add node type E8000 to represent e8000 products. I understand that node type will be deprecated in the future. However, we still need the node type to be added, since Ondatra KNE binding does not work without this (will send another pull for Ondatra). We do not use node type for node registration, so we can remove the added node types if Ondatra does not require that. 
- Add interface mapping for different models of E8000  
- Add unit tests to test the new changes
- Replace the xrd package with Cisco packages that support both node type xrd and e8000.
- Add sample kne cluster using e8000 and xrd nodes


topo/node/cisco/cisco.go
```diff
@@ -148,18 +156,100 @@ func (n *Node) Create(ctx context.Context) error {
...
 }
 
+func constraints(pb *tpb.Node) map[string]string {
+	constraints := map[string]string{
+		"cpu":    "1",
+		"memory": "2Gi",
+	}
+	switch pb.Model {
+	case "8201", "8201-32FH", "8202", "8101-32H", "8102-64H":
+		constraints = map[string]string{
```
marcushines, line 174: why does this increase so much?

topo/node/cisco/cisco.go
```diff
@@ -148,18 +156,100 @@ func (n *Node) Create(ctx context.Context) error {
...
+func generateInterfacesEnvs(pb *tpb.Node) (interfaceList, interfaceMap string) {
+	interfaceList = "MgmtEther0/RP0/CPU0/0"
+	interfaceMap = "MgmtEther0/RP0/CPU0/0:eth0"
+	eths := make([]string, 0)
+	for k, _ := range pb.Interfaces {
+		eths = append(eths, k)
+	}
+	sort.Strings(eths)
+	for _, eth := range eths {
+		if !strings.HasPrefix(eth, "eth") {
```
marcushines, line 196: so what happens to these interfaces? is there an example of them not matching "eth" ?

topo/node/cisco/cisco.go
```diff
@@ -148,18 +156,100 @@ func (n *Node) Create(ctx context.Context) error {
...
+			continue
+		}
+		if pb.Interfaces[eth].Name == "" {
```
marcushines, line 199: i would probably move this remap into a helper function then your for loop is just calling that function for each interface

topo/node/cisco/cisco.go
mojiiba: e8000 containers provide data plane functionalities and matches production image requirement.

topo/node/cisco/cisco.go
mojiiba: sure

topo/node/cisco/cisco.go
```diff
@@ -148,18 +156,100 @@ func (n *Node) Create(ctx context.Context) error {
...
+		}
+	}
+	return constraints
+}
+
+func generateInterfacesEnvs(pb *tpb.Node) (interfaceList, interfaceMap string) {
+	interfaceList = "MgmtEther0/RP0/CPU0/0"
+	interfaceMap = "MgmtEther0/RP0/CPU0/0:eth0"
+	eths := make([]string, 0)
+	for k, _ := range pb.Interfaces {
```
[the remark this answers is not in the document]
mojiiba: Thanks, sure

topo/node/cisco/cisco.go
```diff
@@ -148,18 +156,100 @@ func (n *Node) Create(ctx context.Context) error {
...
+		eths = append(eths, k)
+	}
+	sort.Strings(eths)
+	for _, eth := range eths {
+		if !strings.HasPrefix(eth, "eth") {
+			continue
+		}
+		if pb.Interfaces[eth].Name == "" {
+			ethID, err := strconv.Atoi(strings.TrimPrefix(eth, "eth"))
+			if err != nil {
```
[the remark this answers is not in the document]
mojiiba: sure

topo/node/cisco/cisco.go
mojiiba: e8000 containers provide dataplane functionalities and matches production image requirement.

topo/node/cisco/cisco.go
mojiiba: @marcushines  I need your advice for this case. I did a quick test and it seems that cluster bring up (networkop/init-wait) fails when the interface id is set to anything other than [e|et|eth|...]+[id]. So, I think there is a need to have a  check on KNE to prevent users from setting non-valid ethernet id. However, for this PR, If you agree, I will only handle the interface id in the correct format and logs a warning otherwise.

mojiiba:
@marcushines I have addressed your comment.

topo/node/cisco/cisco.go
```diff
@@ -148,18 +157,112 @@ func (n *Node) Create(ctx context.Context) error {
...
+func getCiscoInterfaceId(model string, ethID int) (string, error) {
+	ciscoInterfacePrefix := ""
+	switch model {
+	case "8201":
+		if ethID-1 <= 23 {
+			ciscoInterfacePrefix = "FourHundredGigE0/0/0/"
+
+		} else if ethID-1 <= 35 {
+			ciscoInterfacePrefix = "HundredGigE0/0/0/"
+		}
```
marcushines, line 197: so what happens for all interfaces outside of that range?
is that an error or ?

topo/node/cisco/cisco.go
marcushines: not changed?

topo/node/cisco/cisco.go
```diff
@@ -148,18 +157,112 @@ func (n *Node) Create(ctx context.Context) error {
...
+	log.Infof("Created Cisco %s node resource %s pod", n.Proto.Model, n.Name())
 	if err := n.CreateService(ctx); err != nil {
 		return err
 	}
-	log.Infof("Created XRD node resource %s services", n.Name())
+	log.Infof("Created Cisco %s node resource %s services", n.Proto.Model, n.Name())
 	return nil
 }
 
+func constraints(pb *tpb.Node) map[string]string {
```
marcushines, line 168: why not properly handle defaults in this function rather then having the if stanza's below?

	constraints := constraints(pb)
	if pb.Constraints == nil {
		pb.Constraints = constraints
	} else {
		if pb.Constraints["cpu"] == "" {
			pb.Constraints["cpu"] = constraints["cpu"]
		}
		if pb.Constraints["memory"] == "" {
			pb.Constraints["memory"] = constraints["memory"]
		}
	}

topo/node/cisco/cisco.go
mojiiba: I changed the code to return an error and fail early similar to the above case.

topo/node/cisco/cisco.go
mojiiba: done

topo/node/cisco/cisco.go
mojiiba: I moved the remap completely to the helper function.

topo/node/cisco/cisco.go
mojiiba: code is changed to return error and fail early. Add ut to check this case as well.

topo/node/cisco/cisco.go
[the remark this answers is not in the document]
mojiiba: this is resolved by returning error and failing safe since this can cause issue for topology bringup that would be hard to debug.

topo/node/cisco/cisco.go
```diff
@@ -148,18 +159,117 @@ func (n *Node) Create(ctx context.Context) error {
...
+	ethWithIdRegx := regexp.MustCompile(`e(t(h(e(r(n(e(t)*)*)*)*)*)*)\d+`) // check for e|et|eth|....
+	ethRegx := regexp.MustCompile(`e(t(h(e(r(n(e(t)*)*)*)*)*)*)`)          // check for e|et|eth|....
+	if !ethWithIdRegx.MatchString(eth) {
+		return "", fmt.Errorf("interface '%s' is invalid", eth)
+	}
+	if pb.Interfaces[eth].Name == "" {
+		// ethWithIdRegx.MatchString(eth) was successfull, so no need to do extra check here
+		ethId, _ := strconv.Atoi(ethRegx.Split(eth, -1)[1])
+		switch pb.Model {
+		case "8201":
```
marcushines, line 204: nit so i would refactor this just a touch

```go
func fmtInt100(eid string) string {
  return fmt.Sprintf("HundredGigE0/0/0/%d", eid)
}

func fmtInt400(eid string) string {
  return fmt.Sprintf("FourHundredGigE0/0/0/%d", eid)
}


if !ethWithIdRegx.MatchString(eth) {
	return "", fmt.Errorf("interface '%s' is invalid", eth)
}
if pb.interface[eth].Name != "" {
  return pb.Interfaces[eth].Name, nil
}
// ethWithIdRegx.MatchString(eth) was successful, so no need to do extra check here
ethId, _ := strconv.Atoi(ethRegx.Split(eth, -1)[1])
eid := ethId - 1 
switch pb.Model {
case "8201":
  switch {
    case eid <= 23:
       return fmtInt400(eid), nil
    case eid <= 35:
       return fmtInt100(eid), nil
    }
   return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth36 is supported on %s ", ethId, pb.Model)
case "8202":
  switch {
    case eid<= 47:
       return fmtInt100(eid), nil
    case eid <= 59:
       return fmtInt400(eid), nil
    case ethId-1 <= 71:
       return fmtInt100(eid), nil
    }
    return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth72 is supported on %s ", ethId, pb.Model)
case "8201-32FH":
  if eid <= 31 {
       return fmtInt400(eid), nil
  }
  return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth32 is supported on %s ", ethId, pb.Model)
case "8101-32H":
  if ethId-1 <= 31 {
    return fmtInt100(eid), nil
  }
  return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth32 is supported on %s ", ethId, pb.Model)
case "8102-64H":
  if ethId-1 <= 63 {
    return fmtInt100(eid), nil
  }
return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth64 is supported on %s ", ethId, pb.Model)
default:
  return fmt.Sprintf("GigabitEthernet0/0/0/%d", eid)
}
```

mojiiba: Thanks a lot! updated as you suggested.

topo/node/cisco/cisco.go
```diff
@@ -190,60 +190,66 @@ func constraints(pb *tpb.Node) *tpb.Node {
...
-				ciscoInterfacePrefix = "HundredGigE0/0/0/"
-			} else {
-				return "", fmt.Errorf("interface id %d can not be mapped to a cisco interface, eth1..eth64 is supported on %s ", ethId, pb.Model)
-			}
-		default:
-			ciscoInterfacePrefix = "GigabitEthernet0/0/0/"
-		}
-		return fmt.Sprintf("%s%d", ciscoInterfacePrefix, ethId-1), nil
-	} else {
+	if !ethWithIdRegx.MatchString(eth) {
```
marcushines: dup

marcushines (approved):

Issue #93, opened 2022-01-23
[ENH] Need support for loading YAML/JSON file for topology


ashutshkumr:
Writing .pb.txt by hand is more error-prone (compared to YAML/JSON), because:
- Adding lists / maps is a little counter-intuitive
- No stable support for linting in popular editors (found some in VSCode marketplace, but they require the format to be .pbtext)
- No online tools available for quick validation / formatting

Moreover, the config in YAML seems to be more compact. Support for JSON is optional.

PR #94, opened 2022-01-23
Added support for loading JSON/YAML files for topology (with .yaml, yml and .json extensions)

ashutshkumr:
Fixes https://github.com/google/kne/issues/93.

File formats are detected based on extensions `.yaml`, `.yml` and `.json`. If no match is found, files are assumed to be .pb.txt (i.e. no check for extension when loading .pb.txt, as was the case before).


ashutshkumr:
Will need to re-create PR using an org account.

PR #98, opened 2022-01-26
Add test coverage for pushFn in cli

marcushines:


cmd/topology/topology_test.go
```diff
@@ -488,3 +496,83 @@ func TestService(t *testing.T) {
 		})
 	}
 }
+
+func TestPush(t *testing.T) {
+	confFile, err := ioutil.TempFile("", "push")
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Fprintln(confFile, "some bytes")
```
shubh90, line 505: remove ?

shubh90 (approved):

cmd/topology/topology_test.go
marcushines: it is generating the test data for the config file load

PR #101, opened 2022-01-27
Update ixia and srl nodes to new pattern

shubh90:


topo/node/srl/srl.go
```diff
@@ -199,15 +206,24 @@ func defaults(pb *topopb.Node) *topopb.Node {
...
 				NodePort: node.GetNextPort(),
 			},
-		},
-		Labels: map[string]string{
+		}
+
+	}
+	if pb.Labels == nil {
+		pb.Labels = map[string]string{
 			"type": topopb.Node_NOKIA_SRL.String(),
```
alexmasi, line 214: I think this is a key that should be added to the map regardless of if the map exists. But do not overwrite an existing value if the key already exists

shubh90: Done

alexmasi (approved):

PR #103, opened 2022-01-27
Update xrd nodes

shubh90:


topo/node/xrd/xrd.go
```diff
@@ -171,13 +172,21 @@ func defaults(pb *tpb.Node) *tpb.Node {
 	if pb.Config.ConfigPath == "" {
 		pb.Config.ConfigPath = "/"
 	}
-	return &tpb.Node{
-		Name: pb.Name,
-		Constraints: map[string]string{
+	if pb.Constraints == nil {
```
alexmasi, line 175: A better pattern is:

if pb.Constraints == nil {
    pb.Constraints = map[string]string{}
}
if _, ok := pb.Constraints["cpu"]; !ok {
    pb.Constraints["cpu"] = "1"
}
if _, ok := pb.Constraints["memory"]; !ok {
    pb.Constraints["memory"] = "2Gi"
}


This also applies to the labels map and the config map

shubh90: I considered it. But since we have this pattern for all nodes I decided to go with it. If we want we can change to this pattern for all nodes in a separate PR ?

alexmasi (approved):

PR #104, opened 2022-01-29
Create cluster

shubh90:


controller/server/main.go
```diff
@@ -0,0 +1,66 @@
...
+	return nil, status.Errorf(codes.Unimplemented, "DeleteCluster RPC method not implemented.")
+}
+
+func (s *server) CreateTopology(ctx context.Context, req *cpb.CreateTopologyRequest) (*cpb.CreateTopologyResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "CreateTopology RPC method not implemented.")
+}
+
+func (s *server) DeleteTopology(ctx context.Context, req *cpb.DeleteTopologyRequest) (*cpb.DeleteTopologyResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "DeleteTopology RPC method not implemented.")
+}
```
alexmasi, line 51: You don't actually need any of these methods, the cpb.UnimplementedTopologyManagerService will automatically return what you've defined here for all un-specified methods

controller/server/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+func main() {
```
alexmasi, line 53: main() should look something like this (the alts creds are required for directpath):

```
func main() {
	flag.Parse()
	addr := fmt.Sprintf(":%d", *port)
	lis, err := net.Listen("tcp6", addr)
	if err != nil {
		log.Fatalf("Failed to listen: %v", err)
	}
	creds := alts.NewServerCreds(alts.DefaultServerOptions())
	s := grpc.NewServer(grpc.Creds(creds))
	gpb.RegisterGNMIServer(s, &server{})
	log.Infof("Server listening at %v", lis.Addr())
	if err := s.Serve(lis); err != nil {
		log.Fatalf("Failed to serve: %v", err)
	}
}
```

shubh90: Done

controller/server/main.go
shubh90: Done

alexmasi (approved):

PR #99, opened 2022-01-27
Add support for passing kubectl down to exec calls

marcushines:

PR #107, opened 2022-02-03
Fix kubectl exec commands

bstoll:
#99 modified the Nokia SRL CLI command with incorrect syntax.  I do not have a CPTX image to test against but it looks like it was modified the same way.


nitinsoniism (approved):
+1 for cptx. Thanks for fixing this.

marcushines (approved):

PR #111, opened 2022-02-09
Implement CreateCluster RPC in controller service + unit tests

shubh90:


controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+	}
+	if err := d.Deploy(ctx, defaultKubeCfg); err != nil {
+		return &cpb.CreateClusterResponse{
+			Name:  req.GetKind().Name,
+			State: cpb.ClusterState_CLUSTER_STATE_ERROR,
+		}, fmt.Errorf("Failed to deploy cluster: %s", err)
+	}
+	log.Infof("Cluster deployed and ready for topology.")
+	return &cpb.CreateClusterResponse{
```
alexmasi, line 106: is it guaranteed to be running? I thought the point of these methods were to return right away and have the user check async for the cluster status to be RUNNING using ClusterInfo

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
-	port = flag.Int("port", 50051, "Controller server port")
+	port           = flag.Int("port", 50051, "Controller server port")
+	defaultKubeCfg = ""
 )
 
 type server struct {
 	cpb.UnimplementedTopologyManagerServer
 }
 
+func getDeployment(req *cpb.CreateClusterRequest) (*deploy.Deployment, error) {
```
alexmasi, line 48: this should be named "newDeployment" instead of get

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+	d := &deploy.Deployment{}
+	switch kind := req.ClusterSpec.(type) {
+	case *cpb.CreateClusterRequest_Kind:
+		k := &deploy.KindSpec{}
```
alexmasi, line 52: simpler to write:

d.Cluster = &deploy.KindSpec{
Name: req.GetKind().Name,
...
}

Also, I think there are fields missing here (specifically the GoogleArtifactRegistries field at least)

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+		k.Version = req.GetKind().Version
+		k.Image = req.GetKind().Image
+		d.Cluster = k
+	default:
+		return nil, fmt.Errorf("cluster type not supported: %T", kind)
+	}
+	switch metallb := req.IngressSpec.(type) {
+	case *cpb.CreateClusterRequest_Metallb:
+		l := &deploy.MetalLBSpec{}
+		if req.GetMetallb().ManifestDir == "" {
```
alexmasi, line 64: cleaner here to write:

l.ManifestDir = req.GetMetallb().ManifestDir
if l.ManifestDir == "" {
  l.ManifestDir = defaultMetallbManifestDir
}

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+			l.ManifestDir = defaultMetallbManifestDir
+		} else {
+			l.ManifestDir = req.GetMetallb().ManifestDir
```
alexmasi, line 67: There should be a check here that the given manifestDir exists and if so translate it to an absolute path, if not throw an error right away

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+		l.IPCount = int(req.GetMetallb().IpCount)
+		d.Ingress = l
+	default:
+		return nil, fmt.Errorf("ingress spec not supported: %T", metallb)
+	}
+	switch meshnet := req.CniSpec.(type) {
+	case *cpb.CreateClusterRequest_Meshnet:
+		m := &deploy.MeshnetSpec{}
+		if req.GetMeshnet().ManifestDir == "" {
+			m.ManifestDir = defaultMeshnetManifestDir
```
alexmasi, line 78: Same as metallb comments

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+	"context"
 	"flag"
 	"fmt"
-	"log"
 	"net"
+	"path/filepath"
 
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/homedir"
+
```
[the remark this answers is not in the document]
shubh90: done. Although my editor vs code automatically adds new line when it formats.

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
 	cpb "github.com/google/kne/proto/controller"
-	"google.golang.org/grpc"
+
 	"google.golang.org/grpc/credentials/alts"
+
+	"google.golang.org/grpc"
+)
+
+const (
+	defaultMetallbManifestDir = "../../manifests/metallb"
```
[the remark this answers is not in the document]
shubh90: Done

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+
+}
+
+func (s *server) CreateCluster(ctx context.Context, req *cpb.CreateClusterRequest) (*cpb.CreateClusterResponse, error) {
+	log.Infof("Creating new cluster")
+	d, err := getDeployment(req)
+	if err != nil {
+		return nil, err
+	}
+	if home := homedir.HomeDir(); home != "" {
```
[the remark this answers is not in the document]
shubh90: Done

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+	}
+	if err := d.Deploy(ctx, defaultKubeCfg); err != nil {
+		return &cpb.CreateClusterResponse{
+			Name:  req.GetKind().Name,
+			State: cpb.ClusterState_CLUSTER_STATE_ERROR,
+		}, fmt.Errorf("Failed to deploy cluster: %s", err)
```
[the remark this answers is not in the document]
shubh90: Done

controller/server/main.go
```diff
@@ -14,24 +14,101 @@
...
+	}
+	log.Infof("Cluster deployed and ready for topology.")
```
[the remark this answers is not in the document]
shubh90: Done

controller/server/main.go
shubh90: Done

controller/server/main.go
shubh90: Done and added missing fields

controller/server/main.go
shubh90: Done

controller/server/main.go
shubh90: Done

controller/server/main.go
shubh90: Yes, it would be running when the call returns. We discussed about this and losing connection is an edge case. If we encounter issues we can change it to return right away. But for now lets not make this complicated with polling on client side.

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
+	if _, err := os.Stat(path); err != nil {
+		return "", fmt.Errorf("path %q does not exist", path)
+	}
+	return path, nil
+}
+
+func Init() {
+	if home := homedir.HomeDir(); home != "" {
+		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+		defaultMeshnetManifestDir = filepath.Join(home, "kne/manifests/meshnet/base")
```
alexmasi, line 132: Join with spaces between just like kubecfg

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to evaluate absolute path: %q", path)
+	}
+	if _, err := os.Stat(path); err != nil {
+		return "", fmt.Errorf("path %q does not exist", path)
+	}
+	return path, nil
+}
+
+func Init() {
```
alexmasi, line 129: This should be called func init() {}

then it will be automatically called and you don't need to call it in main.

Also its customary for this to be at the top of the file, in this case right beneath the first var block

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
+	if home := homedir.HomeDir(); home != "" {
+		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+		defaultMeshnetManifestDir = filepath.Join(home, "kne/manifests/meshnet/base")
+		defaultMetallbManifestDir = filepath.Join(home, "kne/manifests/metallb")
+	}
+
```
alexmasi, line 135: There are extra newlines here and elsewhere that are uneccesaary

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
+			Retain:                   req.GetKind().Retain,
+			GoogleArtifactRegistries: req.GetKind().GoogleArtifactRegistries,
+		}
+	default:
+		return nil, fmt.Errorf("cluster type not supported: %T", kind)
+	}
+	switch metallb := req.IngressSpec.(type) {
+	case *cpb.CreateClusterRequest_Metallb:
+		l := &deploy.MetalLBSpec{}
+		if req.GetMetallb().ManifestDir != "" {
```
alexmasi, line 62: overwriting default is not a good idea, especially with a long running server, follow this pattern instead:

```
path = defaultMetallbManifestDir
if req.GetMetallb().ManifestDir != "" {
    path = req.GetMetallb().ManifestDir
}
p, err := validatePath(path)
if err != nil {
    return nil, fmt.Errorf("failed to validate manifest path %q", path)
}
l.ManifestDir = p
```

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
+			defaultMetallbManifestDir = req.GetMetallb().ManifestDir
+
```
alexmasi, line 64: remove extra newline

controller/server/main.go
```diff
@@ -14,25 +14,129 @@
...
-	"log"
 	"net"
+	"os"
+	"path/filepath"
 
+	"github.com/google/kne/deploy"
 	cpb "github.com/google/kne/proto/controller"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/homedir"
+
```
alexmasi, line 28: remove extra newline

controller/server/main.go
alexmasi: https://golangdocs.com/init-function-in-golang

controller/server/main.go
shubh90: done

controller/server/main.go
shubh90: done

controller/server/main.go
shubh90: done

controller/server/main.go
shubh90: done

controller/server/main.go
shubh90: done

controller/server/main.go
shubh90: done

alexmasi (approved):

PR #114, opened 2022-02-11
Extract the logics of create/delete topology into topo.go

guoshiuan:
This is the first step of refactoring kne_cli library by making
the code reusable for the Topology manager service.

Test
==
* Unit tests are added.
* Compiled and ran the refactored kne_cli in the GCE VM
  (with the order of create/show/delete), compared its output
  with the same parameters/steps with current kne_cli, and
  the output were consistent.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -130,7 +130,6 @@ func createFn(cmd *cobra.Command, args []string) error {
 		Kubecfg:        kubecfg,
 		TopoNewFunc:    topo.New,
 		TopoNewOptions: []topo.Option{topo.WithBasePath(bp)},
-		TopoLoadFunc:   topo.Load,
 		Timeout:        timeout,
 		DryRun:         dryrun,
 	}
@@ -142,7 +141,6 @@ func deleteFn(cmd *cobra.Command, args []string) error {
 		TopoName:     args[0],
 		Kubecfg:      kubecfg,
 		TopoNewFunc:  topo.New,
-		TopoLoadFunc: topo.Load,
 	}
 	return topo.DeleteTopology(cmd.Context(), p)
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -21,7 +21,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/google/kne/cmd/deploy"
 	"github.com/google/kne/cmd/topology"
 	"github.com/google/kne/topo"
@@ -114,6 +113,7 @@ func validateTopology(cmd *cobra.Command, args []string) error {
 
 var (
 	topoNew = topo.New
+	topoLoad = topo.Load
 )
 
 func fileRelative(p string) (string, error) {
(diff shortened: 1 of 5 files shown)


topo/topo.go
```diff
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/google/kne/topo/node"
+  "github.com/golang/protobuf/proto"
```
alexmasi, line 27: run `go fmt ./...` on these changes to format things like this

topo/topo_test.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	tfake "github.com/google/kne/api/clientset/v1beta1/fake"
+	tpb "github.com/google/kne/proto/topo"
+	"github.com/h-fam/errdiff"
+	kfake "k8s.io/client-go/kubernetes/fake"
+
```
alexmasi, line 26: remove extra newline

topo/topo_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/client-go/rest"
+)
+
+var (
+	opts []Option
```
alexmasi, line 31: I don't think this is necessary, the fact that it is global and it gets a reset deferred after a single test function is overly complicated.

Its probably sufficient to instead change the block at the top of the test function to:

```
tf, err := tfake.NewSimpleClientset()
if err != nil {
    t.Fatalf("cannot create fake topology clientset")
}
opts := []Option{
    WithClusterConfig(&rest.Config{}),
    WithKubeClient(kfake.NewSimpleClientset()),
    WithTopoClient(tf),
}
```

topo/topo_test.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+func TestCreateDeletTopology(t *testing.T) {
```
alexmasi, line 34: Better to break this into two test funcs for clarity:

func TestCreateTopology() {}

func TestDeleteTopology() {}

topo/topo_test.go
```diff
@@ -0,0 +1,135 @@
...
+			},
+			TopoNewOptions: opts,
+			DryRun:         true,
+		},
+		testFunc: DeleteTopology,
+		wantErr:  "invalid topology",
+	},
+	}
+
+	for _, tc := range tests {
```
alexmasi, line 129: Use t.Run() for the subtests: go/go-style/decisions#subtests

cmd/root.go
```diff
@@ -114,6 +113,7 @@ func validateTopology(cmd *cobra.Command, args []string) error {
 
 var (
 	topoNew = topo.New
```
alexmasi, line 115: I don't think this stubbing is needed anymore, its not used in tests either. So probably best to streamline the params to not need a load/new func

cmd/root.go
```diff
@@ -125,68 +125,31 @@ func fileRelative(p string) (string, error) {
...
-		return fmt.Errorf("%s: %w", cmd.Use, err)
-	}
-	fmt.Fprintf(out, "Pods:\n")
-	for _, p := range r.Pods {
-		fmt.Fprintf(out, "%s\n", p.Name)
-	}
-
-	return nil
+	log.Infof(bp)
+	p := topo.TopologyParams {
```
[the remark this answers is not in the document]
guoshiuan: The New and Load functions are used to override the topo.New and topo.Load in the unit tests. For the Load function, I removed it since I can use another way to make the testing work.

topo/topo.go
```diff
@@ -445,3 +446,75 @@ func (m *Manager) Node(nodeName string) (node.Node, error) {
 	}
 	return n, nil
 }
+
+// TopologyParams specifies the parameters used by the functions that
+// creates/deletes topology.
+type TopologyParams struct {
+	TopoName       string                                                   // the filename of the topology
+	Kubecfg        string                                                   // the path of kube config
+	TopoLoadFunc   func(fName string) (*tpb.Topology, error)                // the function that returns the topology protobuf from a file
```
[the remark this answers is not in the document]
guoshiuan: Remove these TopoXXXFunc() and only keep the Option.

topo/topo_test.go
```diff
@@ -0,0 +1,131 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package topo
+
+import (
+	"context"
+	// "fmt"
```
alexmasi, line 19: delete these commented imports

alexmasi (commented):
All of my other comments have been addressed, but will leave approval to Marcus after he takes another look

marcushines (approved):

PR #119, opened 2022-02-17
Extract the logics of "show topology service" into topo.go

guoshiuan:
This PR extracts the logics of kne_cli's "show topology service" into GetTopologyServices() in topo.go.

Test
==
- Unit tests are added for both topo/topo_test.go and cmd/topology/topology_test.go.
- Compiled and ran the refactored kne_cli in the GCE VM, compared its output with the same parameters with current kne_cli, and the output were consistent.

### DIFF
--- a/topo/topo.go
+++ b/topo/topo.go
@@ -508,6 +508,26 @@ func (m *Manager) Node(nodeName string) (node.Node, error) {
 	return n, nil
 }
 
+type resourcer interface {
+	Load(context.Context) error
+	Resources(context.Context) (*Resources, error)
+}
+
+// fakeTopology is used for testing only.
+type fakeTopology struct {
+	resources *Resources
+	rErr      error
+	lErr      error
+}
+
+func (f *fakeTopology) Load(context.Context) error {
+	return f.lErr
+}
+
+func (f *fakeTopology) Resources(context.Context) (*Resources, error) {
+	return f.resources, f.rErr
+}
+
 // TopologyParams specifies the parameters used by the functions that
 // creates/deletes/show topology.
 type TopologyParams struct {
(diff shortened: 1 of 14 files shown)


topo/topo.go
```diff
@@ -453,6 +453,26 @@ func (m *Manager) Node(nodeName string) (node.Node, error) {
 	return n, nil
 }
 
+type resourcer interface {
+	Load(context.Context) error
+	Resources(context.Context) (*Resources, error)
+}
+
+// fakeTopology is used for testing only.
+type fakeTopology struct {
```
alexmasi, line 462: you can define this in the _test.go file

edit: didn't see its use in topo params

topo/topo.go
```diff
@@ -461,6 +481,7 @@ type TopologyParams struct {
 	TopoNewOptions []Option // the options used in the TopoNewFunc
 	Timeout        time.Duration
 	DryRun         bool
+	fakeTopo       *fakeTopology // a fake topology manager for testing only
```
alexmasi, line 484: something about this feels strange to me, but it doesn't exactly make sense as a topoNewOption either

topo/topo.go
```diff
@@ -522,3 +543,83 @@ func DeleteTopology(ctx context.Context, params TopologyParams) error {
...
+	topopb, err := Load(params.TopoName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load %s: %+v", params.TopoName, err)
+	}
+
+	var t resourcer
+	if params.fakeTopo != nil {
+		t = params.fakeTopo
+	} else {
+		t, err = New(params.Kubecfg, topopb, params.TopoNewOptions...)
```
alexmasi, line 587: Maybe a solution would be to stub New which should return a resourcer

Then each test case provides a stub for the New func which returns a fakeTopo

I think this is the cleanest solution, but open to discussion

PR #120, opened 2022-02-17
Extract the logics of "show topology service" into topo.go

guoshiuan:
This PR extracts the logics of kne_cli's "show topology service" into GetTopologyServices() in topo.go.

Test
==
* Unit tests are added for both topo/topo_test.go and cmd/topology/topology_test.go.
* Compiled and ran the refactored kne_cli in the GCE VM, compared its output with the same parameters with current kne_cli, and the output were consistent.

### DIFF
--- a/topo/topo.go
+++ b/topo/topo.go
@@ -49,59 +49,17 @@ import (
 	_ "github.com/google/kne/topo/node/srl"
 )
 
-// Resourcer loads and reports resources of a topology.
-type Resourcer interface {
-	Load(context.Context) error
-	Resources(context.Context) (*Resources, error)
-}
-
-// Creator creates a topology.
-type Creator interface {
-	Load(context.Context) error
-	Push(context.Context) error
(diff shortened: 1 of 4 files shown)


topo/topo.go
```diff
@@ -49,6 +49,61 @@ import (
 	_ "github.com/google/kne/topo/node/srl"
 )
 
+// Resourcer loads and reports resources of a topology.
+type Resourcer interface {
```
alexmasi, line 53: all of these separate interfaces are not required it seems, was this done to increase clarity?

It might make more sense to define a single TopologyManager interface like you've done before that includes all these methods inside of it directly. I don't see a real benefit to defining these subinterfaces separately.

What do you think?

guoshiuan: Yes, it is just for clarity purpose. Revised the code by putting methods directly into the TopologyManager interface; we can easily create sub-interface if needed in the future.

alexmasi (approved):

marcushines (approved):

PR #122, opened 2022-02-22
Add topology create/delete/show to topology manager

guoshiuan:
Add topology create/delete/show handler to topology manager.

### DIFF
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -181,15 +181,8 @@ func (s *server) ShowTopology(ctx context.Context, req *cpb.ShowTopologyRequest)
 		TopoName: req.TopologyName,
 		Kubecfg:  defaultKubeCfg,
 	}
-	tpb, state, err := topo.GetTopologyServices(ctx, p)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to show topology: %v", err)
-	}
 
-	return &cpb.ShowTopologyResponse{
-		State:    state,
-		Topology: tpb,
-	}, nil
+	return topo.GetTopologyServices(ctx, p)
 }
 
 func main() {
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -37,8 +37,7 @@ var (
 	defaultMetallbManifestDir = ""
 	defaultMeshnetManifestDir = ""
 	// Flags.
-	port    = flag.Int("port", 50051, "Controller server port")
-	kubecfg string
+	port = flag.Int("port", 50051, "Controller server port")
 )
 
 func init() {
@@ -47,8 +46,6 @@ func init() {
 		defaultMeshnetManifestDir = filepath.Join(home, "kne", "manifests", "meshnet", "base")
 		defaultMetallbManifestDir = filepath.Join(home, "kne", "manifests", "metallb")
 	}
-	// flag.IntVar(&port, "port", 50051, "Controller server port")
-	flag.StringVar(&kubecfg, "kubecfg", defaultKubeCfg, "Kube config")
 }
 
 type server struct {
@@ -153,7 +150,7 @@ func (s *server) CreateTopology(ctx context.Context, req *cpb.CreateTopologyRequ
 	p := topo.TopologyParams{
 		TopoName:       req.Topology.Name,
 		TopoNewOptions: []topo.Option{topo.WithBasePath(filepath.Dir(bp))},
-		Kubecfg:        kubecfg,
+		Kubecfg:        req.Kubecfg,
 	}
 	err = topo.CreateTopology(ctx, p)
 	if err != nil {
@@ -174,7 +171,6 @@ func (s *server) DeleteTopology(ctx context.Context, req *cpb.DeleteTopologyRequ
 	log.Infof("Received DeleteTopology request: %v", req)
 	p := topo.TopologyParams{
 		TopoName: req.TopologyName,
-		Kubecfg:  kubecfg,
 	}
 	err := topo.DeleteTopology(ctx, p)
 	if err != nil {
@@ -189,7 +185,6 @@ func (s *server) ShowTopology(ctx context.Context, req *cpb.ShowTopologyRequest)
 	log.Infof("Received ShowTopology request: %v", req)
 	p := topo.TopologyParams{
 		TopoName: req.TopologyName,
-		Kubecfg:  kubecfg,
 	}
 	tpb, err := topo.GetTopologyServices(ctx, p)
 	if err != nil {
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -24,8 +24,9 @@ import (
 	log "github.com/golang/glog"
 	"github.com/google/kne/deploy"
 	cpb "github.com/google/kne/proto/controller"
+	"github.com/google/kne/topo"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/alts"
+	// "google.golang.org/grpc/credentials/alts"
 	"k8s.io/client-go/util/homedir"
 )
 
@@ -34,7 +35,8 @@ var (
 	defaultMetallbManifestDir = ""
 	defaultMeshnetManifestDir = ""
 	// Flags.
-	port = flag.Int("port", 50051, "Controller server port")
+	port    = flag.Int("port", 50051, "Controller server port")
+	kubecfg string
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


controller/server/main.go
```diff
@@ -138,6 +144,66 @@ func validatePath(path string) (string, error) {
...
+		return nil, status.Errorf(codes.InvalidArgument, "invalid topology path: %v", err)
+	}
+	p := topo.TopologyParams{
+		TopoName:       req.Topology.Name,
+		TopoNewOptions: []topo.Option{topo.WithBasePath(filepath.Dir(bp))},
+		Kubecfg:        kubecfg,
+	}
+	err = topo.CreateTopology(ctx, p)
+	if err != nil {
+		return &cpb.CreateTopologyResponse{
```
marcushines, line 160: err if you return an error here there is no contract on the object returned - if you want data included that the user can "rely on" you should return that as part of the error

controller/server/main.go
```diff
@@ -138,6 +144,66 @@ func validatePath(path string) (string, error) {
...
+func (s *server) CreateTopology(ctx context.Context, req *cpb.CreateTopologyRequest) (*cpb.CreateTopologyResponse, error) {
+	log.Infof("Received CreateTopology request: %v", req)
+	bp, err := validatePath(req.Topology.Name)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid topology path: %v", err)
+	}
+	p := topo.TopologyParams{
+		TopoName:       req.Topology.Name,
+		TopoNewOptions: []topo.Option{topo.WithBasePath(filepath.Dir(bp))},
+		Kubecfg:        kubecfg,
```
alexmasi, line 156: This value should be the kubecfg provided: https://github.com/google/kne/blob/3388d2b60b540d51ad693acff6b5f81fedadd0f6/proto/controller.proto#L126

or the defaultKubeCfg if empty string

controller/server/main.go
```diff
@@ -43,6 +47,8 @@ func init() {
 		defaultMeshnetManifestDir = filepath.Join(home, "kne", "manifests", "meshnet", "base")
 		defaultMetallbManifestDir = filepath.Join(home, "kne", "manifests", "metallb")
 	}
+	// flag.IntVar(&port, "port", 50051, "Controller server port")
+	flag.StringVar(&kubecfg, "kubecfg", defaultKubeCfg, "Kube config")
```
alexmasi, line 51: the kubecfg flag can be removed. it makes sense for the CLI, but does not make sense at server creation time. instead the user passes the kubecfg field in the request for topo creation as I commented down below

controller/server/main.go
guoshiuan: Done. Also revised CreateCluster() based on discussion with Alex.

For a finer granularity of error type, I will have another PR to propagate the error details from the library in topo.go.

controller/server/main.go
```diff
@@ -138,6 +137,59 @@ func validatePath(path string) (string, error) {
...
+	bp, err := validatePath(req.Topology.Name)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid topology path: %v", err)
+	}
+	p := topo.TopologyParams{
+		TopoName:       req.Topology.Name,
+		TopoNewOptions: []topo.Option{topo.WithBasePath(filepath.Dir(bp))},
+		Kubecfg:        req.Kubecfg,
+	}
+	err = topo.CreateTopology(ctx, p)
```
marcushines, line 151: nit

if err := topo.CreateTopology(ctx, p); err != nil { ...}

controller/server/main.go
```diff
@@ -138,6 +137,59 @@ func validatePath(path string) (string, error) {
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to create topology: %v", err)
+	}
+
+	return &cpb.CreateTopologyResponse{
+		TopologyName: req.Topology.GetName(),
+		State:        cpb.TopologyState_TOPOLOGY_STATE_RUNNING,
```
marcushines, line 158: should this state be stored? i.e. is thera some other get call which would return this status?

controller/server/main.go
```diff
@@ -138,6 +137,59 @@ func validatePath(path string) (string, error) {
...
+	}, nil
+
+}
+
+func (s *server) DeleteTopology(ctx context.Context, req *cpb.DeleteTopologyRequest) (*cpb.DeleteTopologyResponse, error) {
+	log.Infof("Received DeleteTopology request: %v", req)
+	p := topo.TopologyParams{
+		TopoName: req.TopologyName,
+	}
+	err := topo.DeleteTopology(ctx, p)
```
marcushines, line 168: same as above on nit for if structure

controller/server/main.go
```diff
@@ -138,6 +137,59 @@ func validatePath(path string) (string, error) {
...
+	p := topo.TopologyParams{
+		TopoName: req.TopologyName,
+	}
+	tpb, err := topo.GetTopologyServices(ctx, p)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to show topology: %v", err)
+	}
+
+	return &cpb.ShowTopologyResponse{
+		State:    cpb.TopologyState_TOPOLOGY_STATE_RUNNING,
```
marcushines, line 188: so this feels like we are giving a state which we didn't actually check we just assume?

alexmasi: agreed, there needs to be some code here that can check the status of a topology and classify which of the enum states the topo is in

controller/server/main.go
guoshiuan: The latest state can be got via ShowTopology() call.

controller/server/main.go
guoshiuan: Added a check to derive the topology state with the node's pod state.

topo/topo.go
```diff
@@ -569,25 +570,62 @@ var (
...
+	}
+	if cntTable[corev1.PodPending] > 0 {
+		return cpb.TopologyState_TOPOLOGY_STATE_CREATING
+	}
+	return cpb.TopologyState_TOPOLOGY_STATE_UNKNOWN
+}
+
 // GetTopologyServices returns the topology information.
-func GetTopologyServices(ctx context.Context, params TopologyParams) (*tpb.Topology, error) {
+func GetTopologyServices(ctx context.Context, params TopologyParams) (*tpb.Topology, cpb.TopologyState, error) {
```
marcushines, line 611: why add a another return value here?
just return a cpb.ShowTopologyReponse?

since you are returning the cpb.Topologystate now you are already having that proto as a dependency

guoshiuan: GetTopologyServices() is a generic library and is also used by kne_cli's "topology services" command.

marcushines: right but you are making 3 return values when you don't need to

you can just return a *cpb.ShowTopologyResponse, error

and still end up with the same dependencies

guoshiuan: Done.

topo/topo.go
```diff
@@ -569,25 +570,70 @@ var (
...
+	return cpb.TopologyState_TOPOLOGY_STATE_UNKNOWN
+}
+
 // GetTopologyServices returns the topology information.
-func GetTopologyServices(ctx context.Context, params TopologyParams) (*tpb.Topology, error) {
+func GetTopologyServices(ctx context.Context, params TopologyParams) (*cpb.ShowTopologyResponse, error) {
 	topopb, err := Load(params.TopoName)
 	if err != nil {
-		return nil, fmt.Errorf("failed to load %s: %+v", params.TopoName, err)
+		return &cpb.ShowTopologyResponse{
```
marcushines, line 614: why are you returning both an object and an error?

controller/server/main.go
```diff
@@ -138,6 +137,54 @@ func validatePath(path string) (string, error) {
...
+}
+
+func (s *server) ShowTopology(ctx context.Context, req *cpb.ShowTopologyRequest) (*cpb.ShowTopologyResponse, error) {
+	log.Infof("Received ShowTopology request: %v", req)
+	p := topo.TopologyParams{
+		TopoName: req.TopologyName,
+		Kubecfg:  defaultKubeCfg,
+	}
+
+	return topo.GetTopologyServices(ctx, p)
```
marcushines, line 185: resp, err := topo.GetTopologyServices(ctx, topo.TopologyParams{
		TopoName: req.TopologyName,
		Kubecfg:  defaultKubeCfg,
	})
if err != nil {
  return nil, status.Errorf(".....)
 }
 return resp,nil

topo/topo.go
guoshiuan: oops... done.

marcushines (approved):

PR #123, opened 2022-02-23
Add ShowCluster method implementation to TopologyManager

alexmasi:
- generalize DeleteCluster to work with non-kind clusters
- ShowCluster status check works for all cluster, cni, and ingress implementations
- topology_names repeated field is currently not used, unclear how to provide this list. will add in a later PR if needed
- Delete/ShowCluster only works for clusters created using CreateCluster. This is due to the extra deployment info stored during cluster creation such as ingress/cni/cluster type in order to perform relevant health checks and deletion

### DIFF
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -51,7 +51,7 @@ func init() {
 type server struct {
 	cpb.UnimplementedTopologyManagerServer
 
-	mu sync.Mutex
+	mu          sync.Mutex
 	deployments map[string]*deploy.Deployment
 }
 
@@ -116,7 +116,7 @@ func newDeployment(req *cpb.CreateClusterRequest) (*deploy.Deployment, error) {
 }
 
 func (s *server) CreateCluster(ctx context.Context, req *cpb.CreateClusterRequest) (*cpb.CreateClusterResponse, error) {
-	log.Infof("Received CreateCluster request: %+v", req)
+	log.Infof("Received CreateCluster request: %v", req)
 	d, err := newDeployment(req)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "unable to parse request: %v", err)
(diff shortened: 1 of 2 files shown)


controller/server/main.go
```diff
@@ -122,61 +116,60 @@ func newDeployment(req *cpb.CreateClusterRequest) (*deploy.Deployment, error) {
...
 	}
 	log.Infof("Parsed request into deployment: %v", d)
 	if err := d.Deploy(ctx, defaultKubeCfg); err != nil {
 		resp := &cpb.CreateClusterResponse{
-			Name:  req.GetKind().Name,
+			Name:  d.Cluster.GetName(),
 			State: cpb.ClusterState_CLUSTER_STATE_ERROR,
 		}
 		return resp, status.Errorf(codes.Internal, "failed to deploy cluster: %v", err)
 	}
```
guoshiuan, line 131: Should it be nil since there is an error?

controller/server/main.go
```diff
@@ -122,61 +116,60 @@ func newDeployment(req *cpb.CreateClusterRequest) (*deploy.Deployment, error) {
...
-	log.Infof("Cluster %q deployed and ready for topology", req.GetKind().Name)
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.deployments[d.Cluster.GetName()] = d
```
guoshiuan, line 134: Do we need to check whether the cluster exists already?

controller/server/main.go
alexmasi: yeah I planned to pick up these changes from your PR that is approved

controller/server/main.go
alexmasi: Good catch

guoshiuan (approved):

PR #124, opened 2022-03-01
Use WithTopology() to create/delete/show topology in the Topology Manager

guoshiuan:
This PR also adds the following features:
- Validate the topology protobuf in the request, e.g., whether router config path is valid.
- Revise the unit test for topo.go

### DIFF
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -25,16 +25,19 @@ import (
 	log "github.com/golang/glog"
 	"github.com/google/kne/deploy"
 	cpb "github.com/google/kne/proto/controller"
+	tpb "github.com/google/kne/proto/topo"
 	"github.com/google/kne/topo"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/alts"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/prototext"
 	"k8s.io/client-go/util/homedir"
 )
 
 var (
 	defaultKubeCfg            = ""
+	defaultTopoBasePath       = ""
 	defaultMetallbManifestDir = ""
 	defaultMeshnetManifestDir = ""
 	// Flags.
@@ -44,6 +47,7 @@ var (
 func init() {
 	if home := homedir.HomeDir(); home != "" {
 		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+		defaultTopoBasePath = filepath.Join(home, "kne-internal", "examples")
 		defaultMeshnetManifestDir = filepath.Join(home, "kne", "manifests", "meshnet", "base")
 		defaultMetallbManifestDir = filepath.Join(home, "kne", "manifests", "metallb")
 	}
@@ -52,12 +56,17 @@ func init() {
 type server struct {
 	cpb.UnimplementedTopologyManagerServer
 
-	mu          sync.Mutex
+	muDeploy    sync.Mutex // guards deployements map
 	deployments map[string]*deploy.Deployment
+	muTopo      sync.Mutex        // guards topos map
+	topos       map[string][]byte // stores the topology protobuf from the initial topology creation request
 }
 
 func newServer() *server {
-	return &server{deployments: map[string]*deploy.Deployment{}}
+	return &server{
+		deployments: map[string]*deploy.Deployment{},
+		topos:       map[string][]byte{},
+	}
 }
 
 func newDeployment(req *cpb.CreateClusterRequest) (*deploy.Deployment, error) {
@@ -123,8 +132,8 @@ func (s *server) CreateCluster(ctx context.Context, req *cpb.CreateClusterReques
 		return nil, status.Errorf(codes.InvalidArgument, "unable to parse request: %v", err)
 	}
 	log.Infof("Parsed request into deployment: %v", d)
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	s.muDeploy.Lock()
+	defer s.muDeploy.Unlock()
 	if _, ok := s.deployments[d.Cluster.GetName()]; ok { // if OK
 		return nil, status.Errorf(codes.AlreadyExists, "cluster %q already exists", d.Cluster.GetName())
 	}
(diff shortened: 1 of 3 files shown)


controller/server/main.go
```diff
@@ -44,6 +47,7 @@ var (
 func init() {
 	if home := homedir.HomeDir(); home != "" {
 		defaultKubeCfg = filepath.Join(home, ".kube", "config")
+		defaultTopoBasePath = filepath.Join(home, "kne-internal", "examples")
```
alexmasi, line 50: Should use "kne" instead of "kne-internal" here as the default

We should not make the assumption that the user has kne-internal cloned and instead rely on "kne".

For our use case I'll specify the full path of the config

controller/server/main.go
```diff
@@ -52,12 +56,17 @@ func init() {
 type server struct {
 	cpb.UnimplementedTopologyManagerServer
 
-	mu          sync.Mutex
+	muDeploy    sync.Mutex // guards deployements map
 	deployments map[string]*deploy.Deployment
+	muTopo      sync.Mutex        // guards topos map
+	topos       map[string][]byte // stores the topology protobuf from the initial topology creation request
```
alexmasi, line 62: Is there a reason to store the raw bytes instead of having the map be of type map[string]*tpb.Topology ?

controller/server/main.go
```diff
@@ -183,35 +192,73 @@ func validatePath(path string) (string, error) {
...
+		return nil, status.Errorf(codes.AlreadyExists, "topology %q already exists", req.Topology.GetName())
+	}
+
+	log.Infof("Validating the topology protobuf...")
+	for _, node := range topoPb.Nodes {
+		if node.GetType() == tpb.Node_IXIA_TG {
+			continue
+		}
+		if node.GetConfig() == nil {
+			return nil, status.Errorf(codes.NotFound, "missing config field for node %q", node.GetName())
```
alexmasi, line 215: above we are returning codes.InvalidArgument for when a specified file is not found.

Which code do you think is better for this?

I would lean towards invalid argument here

controller/server/main.go
```diff
@@ -183,35 +192,73 @@ func validatePath(path string) (string, error) {
...
+		}
+		if node.GetConfig().GetFile() == "" {
+			return nil, status.Errorf(codes.NotFound, "config file not specified for node %q", node.GetName())
```
alexmasi, line 218: Also a config section is not required (you are allowed to bring up a topology with no initial config), so it should fail here

It should only fail if a config file path is provided and it isn't valid

topo/topo.go
```diff
@@ -116,10 +117,6 @@ func WithBasePath(s string) Option {
 
 // New creates a new topology manager based on the provided kubecfg and topology.
 func New(kubecfg string, pb *tpb.Topology, opts ...Option) (TopologyManager, error) {
```
alexmasi, line 119: what is the pb argument used in this func for?

instead we are specifying potentially a different pb  with options?

controller/server/main.go
guoshiuan: Done. Also revised the code so it won't add this base path when checking a config file.

controller/server/main.go
guoshiuan: After checking https://grpc.github.io/grpc/core/md_doc_statuscodes.html, I think invalid argument makes more sense. Done.

controller/server/main.go
guoshiuan: Thanks! Revised the code.

topo/topo.go
guoshiuan: Yes, the pb specified will be replaced if WithTopology() is in the Option.

This New() with non-nil pb is still used by KNE cli commands. I will have another PR to revise those code and remove pb from the function signature to avoid confusion..

controller/server/main.go
guoshiuan: In current implementation of Create/Delete/Show topology APIs, the input *tpb.Topology will be modified after the call is done. i.e., this topology pb will be modified by the TopologyManager's Load() method. 

Several ways to fix this problem:
1) Change the current API implementation: instead of getting and loading a protobuf (topo.Load() and then calling manager.Load()), we can passed a "loaded" topology protobuf.
2) Without changing the current API implementation, store it as a *tpb.Topology and the callers need to have a deep copy of it when using it.

Storing it as raw bytes is the simplest way to work around this issue... Any thoughts on this?

topo/topo.go
alexmasi: Cool, can you file a cleanup task in the backlog for this additional work?

controller/server/main.go
```diff
@@ -183,35 +192,69 @@ func validatePath(path string) (string, error) {
...
+		return nil, status.Errorf(codes.InvalidArgument, "missing topology name")
+	}
+
+	s.muTopo.Lock()
+	defer s.muTopo.Unlock()
+	if _, ok := s.topos[topoPb.GetName()]; ok {
+		return nil, status.Errorf(codes.AlreadyExists, "topology %q already exists", req.Topology.GetName())
+	}
+
+	log.Infof("Validating the topology protobuf...")
```
alexmasi, line 209: change this log message to indicate that it is only checking the file paths for validity, not the entire proto

alexmasi: looking more at this I think the message can just be entirely removed given the follow up "checking config path" logging if it comes up

controller/server/main.go
```diff
@@ -183,35 +192,69 @@ func validatePath(path string) (string, error) {
...
+	// Saves the original topology protobuf.
+	txtPb, err := prototext.Marshal(topoPb)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "invalid topology protobuf: %v", err)
 	}
 
-	if err = topo.CreateTopology(ctx, p); err != nil {
+	if err := topo.CreateTopology(ctx, topo.TopologyParams{
+		TopoNewOptions: []topo.Option{topo.WithTopology(topoPb)},
+		Kubecfg:        req.Kubecfg,
```
alexmasi, line 229: this path needs to validated and use the default value if not valid

topo/topo.go
```diff
@@ -629,9 +643,13 @@ func (s *sMap) TopoState() cpb.TopologyState {
 
 // GetTopologyServices returns the topology information.
 func GetTopologyServices(ctx context.Context, params TopologyParams) (*cpb.ShowTopologyResponse, error) {
-	topopb, err := Load(params.TopoName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load %s: %+v", params.TopoName, err)
+	var topopb *tpb.Topology
+	var err error
+	if params.TopoName != "" {
+		topopb, err = Load(params.TopoName)
```
alexmasi, line 649: So this doesn't refer to the name of the topo, but rather the topo file path?

controller/server/main.go
alexmasi: Lets discuss this at our meeting today

controller/server/main.go
guoshiuan: Done

controller/server/main.go
guoshiuan: Done.

topo/topo.go
guoshiuan: It is referring to file path. I can have another PR to avoid the confusion.

alexmasi (approved):

PR #125, opened 2022-03-02
Fix kubecfg validation

alexmasi:
- also use `defaultTopoBasePath` if the config path provided isn't absolute

### DIFF
--- a/controller/server/main.go
+++ b/controller/server/main.go
@@ -203,11 +203,11 @@ func (s *server) CreateTopology(ctx context.Context, req *cpb.CreateTopologyRequ
 		if !filepath.IsAbs(path) {
 			path = filepath.Join(defaultTopoBasePath, path)
 		}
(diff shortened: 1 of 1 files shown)


controller/server/main.go
```diff
@@ -212,20 +199,28 @@ func (s *server) CreateTopology(ctx context.Context, req *cpb.CreateTopologyRequ
 			// topology with no initial config.
 			continue
 		}
-		log.Infof("Check config path: %q", node.GetConfig().GetFile())
+		path := node.GetConfig().GetFile()
+		if !filepath.IsAbs(path) {
+			path = filepath.Join(defaultTopoBasePath, path)
+		}
+		node.GetConfig().ConfigData = &tpb.Config_File{File: path}
```
guoshiuan, line 206: Should we set the ConfigData (line 206) after the path is validated (line 210)?

guoshiuan (approved):

PR #127, opened 2022-03-10
update metallb deployment to use pinned version

marcushines:


manifests/metallb/metallb.yaml
```diff
@@ -364,7 +364,7 @@ spec:
             secretKeyRef:
               name: memberlist
               key: secretkey
-        image: quay.io/metallb/speaker:main
+        image: quay.io/metallb/speaker:v0.12.1
```
guoshiuan, line 367: Is this a temporary change to work around some issue?

alexmasi:
I was just trying this out, about to send a PR. But then I tried main and I couldn't repro the issue. Did you see the issue? Also we should change both the speaker and controller to be pinned

manifests/metallb/metallb.yaml
leomermelstein: metallib:main (apparently) doesn't accept '-config' anymore, breaking this deployment. We need to controller:main also needs to be changed to controller:v0.12.1 (at least until someone can sort out what happened to metallib)

marcushines: updated both images

leomermelstein:
Please change controller:main to controller:v0.12.1 also, as that seems to be broken in the same way. Thanks.

alexmasi (approved):

guoshiuan (approved):

PR #130, opened 2022-03-24
KNE node refactor for nodes backed by more than one pods/services

ashutshkumr:
Creating this as a draft pull request because:
- Need to check how google/kne repository CI may use private image of ixia-c-operator (if it indeed uses it for any E2E tests)
- Need a quick overview of major changes done overall (so I can add more detailed unit tests for newly introduced functions)

Major changes:
- Each node is now considered to be capable of deployed as multiple pods / services
  - Implication being, querying pods or services for each node may return more than one items
- IxiaTG node handler has been rewritten to interface with new ixia-c-operator (containing OTG node refactor changes)


alexmasi:
The automated CI does not yet test anything with traffic (although we would very much like to)

For any end-to-end tests we would be using Ondatra, which does not yet support traffic tests

marcushines (commented):
What version ixia-c should we be testing against? the current release?

ashutshkumr:
@marcushines please use `ixiacom/ixia-c-operator:0.1.77`, which will automatically use ixia-c builds from latest release https://github.com/open-traffic-generator/ixia-c/releases/download/v0.0.1-2738/ixia-configmap.yaml

Kindly let me know if you need help with anything.

ashutshkumr:
@marcushines regarding the backwards compatibility, which one of following are you suggesting ?
- `newer` ixia-c-operator should be able to deploy `new` KNE topology with `older` ixia-c builds. 
- `newer` ixia-c-operator should be able to deploy `old` KNE topology with `older` ixia-c builds.

The challenge with first approach is that we also had to make changes to `ixia-c-controller` and `ixia-c-protocol-engine` to:
- Accept interface names for port locations (e.g. `eth1`) instead of long addr string (e.g. `localhost:5555+localhost:50071`).
- Bind `ixia-c-protocol-engine` to interface with names other than `eth1` (earlier it would only bind to `eth1` interface)

Please note however, `older` ixia-c-operator will still work with both `newer` and `older` ixia-c builds.

ashutshkumr:
further tracked by https://github.com/google/kne/pull/133

PR #147, opened 2022-05-10
Update arista node in 2node-ixia-ceos to use startup config and self-signed certs.

greg-dennis:


alexmasi:
Consider changing the 2node-ixia-ceos-ifc-group.pb.txt and 2node-ixia-ceos.yaml as well as those are intended to be the same (or small modifications) of the file you modified

alexmasi (approved):

PR #150, opened 2022-05-13
Standardize dependency checking for kne deploy

alexmasi:
Before the dependency checking was not uniform, some methods checked for required binaries and others didn't.

Now "docker" and "kubectl" are considered dependencies to make any deployment

"kind" is a dependency to make deployments with kind cluster

"gcloud" is only a dependency when using kind linked with GAR

Other changes:
- removed kind provider deployment without kind cli (now that "kind" cli is a hard requirement, also some of the optional fields like gar support, container loading, deletion etc. were only compatible with cli deployment)
- recycle is now checked using kubectl instead of the kind provider
- clean up code in tests
- go fmt all of kne

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -106,10 +106,9 @@ func (d *Deployment) String() string {
 	return string(b)
 }
 
-func checkDependencies() error {
-	bins := []string{"docker", "kubectl"}
+func (d *Deployment) checkDependencies() error {
 	var errs errlist.List
-	for _, bin := range bins {
+	for _, bin := range []string{"docker", "kubectl"} {
 		if _, err := execLookPath(bin); err != nil {
 			errs.Add(fmt.Errorf("install dependency %q to deploy", bin))
 		}
@@ -118,7 +117,7 @@ func checkDependencies() error {
 }
 
 func (d *Deployment) Deploy(ctx context.Context, kubecfg string) error {
-	if err := checkDependencies(); err != nil {
+	if err := d.checkDependencies(); err != nil {
 		return err
 	}
 	log.Infof("Deploying cluster...")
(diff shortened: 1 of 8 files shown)


deploy/deploy.go
```diff
@@ -117,13 +106,27 @@ func (d *Deployment) String() string {
 	return string(b)
 }
 
+func checkDependencies() error {
+	bins := []string{"docker", "kubectl"}
+	var errs errlist.List
+	for _, bin := range bins {
```
marcushines, line 112: for _, bin := []string{"docker", "kubectl"}

deploy/deploy.go
```diff
@@ -215,48 +218,20 @@ type KindSpec struct {
...
-		if err != nil {
-			return err
-		}
-		for _, v := range clusters {
-			if k.Name == v {
-				log.Infof("Recycling existing cluster: %s", v)
-				return nil
-			}
-		}
+	if _, err := execLookPath("kind"); err != nil {
```
marcushines, line 226: maybe we should codify for each spec a "depends" on func like you did above but attach it to the struct?

deploy/deploy.go
```diff
@@ -215,48 +218,20 @@ type KindSpec struct {
...
-			cluster.CreateWithRetain(k.Retain),
-			cluster.CreateWithWaitForReady(k.Wait),
-			cluster.CreateWithKubeconfigPath(k.Kubecfg),
-			cluster.CreateWithDisplayUsage(true),
-			cluster.CreateWithDisplaySalutation(true),
-		); err != nil {
-			return errors.Wrap(err, "failed to create cluster using kind client")
+	if k.Recycle {
+		log.Infof("Attempting to recycle existing cluster %q...", k.Name)
+		if err := execer.Exec("kubectl", "cluster-info", "--context", fmt.Sprintf("kind-%s", k.Name)); err == nil {
```
marcushines, line 231: should you have a dep check on kubectl? - like is there a guarantee that this order is always maintained?

marcushines (dismissed):

deploy/deploy.go
alexmasi: I see two ways of doing it:

1) having each "Spec" list out the dependencies and call checkDependencies for each one before any method

2) having docker and kubectl be dependencies of any deployment and have the spec dependencies only include the remainder of the dependencies (this is the standard that this PR implements)

I can see potentially a 3rd design that steals from (1) and (2) where docker and kubectl are checked for a deployment regardless of the specs and spec dependencies are checked before ever method call, not just deploy.

I think the 3rd option is the best considering these specs should not be called outside of Deployment.Deploy() and are only exported due to yaml marshaling

deploy/deploy.go
alexmasi: yeah I agree, we can have a depends for the spec, additionally options such as googleArtifactRegistry access requires an additional dep (gcloud) which we could check in the depends func based on the optional fields provided

deploy/deploy.go
alexmasi: done, will push a commit soon

deploy/deploy.go
alexmasi: done

deploy/deploy.go
alexmasi: done

marcushines (approved):

PR #158, opened 2022-06-28
Support KNE deployment with kind with bridge rather than ptp

marcushines:
This adds support for deploying a kind cluster with the bridge
CNI rather than the ptp CNI.
For network devices whose management interfaces are normally
configured as routable interfaces the PTP interface type
has problems when mgmt interfaces routes overlap with assigned
address space


deploy/deploy.go
```diff
@@ -262,10 +264,20 @@ func (k *KindSpec) Deploy(ctx context.Context) error {
...
+	}
+	log.Infof("Creating kind cluster with: %s", args)
 	if err := execer.Exec("kind", args...); err != nil {
 		return errors.Wrap(err, "failed to create cluster")
 	}
 	log.Infof("Deployed kind cluster: %s", k.Name)
+	for _, s := range k.AdditionalManifests {
+		log.Infof("Found manifest %q", s)
+		if err := execer.Exec("kubectl", "apply", "-f", s); err != nil {
+			return errors.Wrap(err, "failed to deploy kindnet bridge")
```
guoshiuan, line 278: Is AdditionalManifests only for kind net bridge?

manifests/kind/kind-bridge.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - name: install-cni-bin
+        image: aojea/kindnetd:v1.0.0
+        command: ['sh', '-c', 'cp -f /opt/cni/bin/* /cni/ ']
+        volumeMounts:
+        - name: cni-bin
+          mountPath: /cni
+      containers:
+      - name: kindnet-cni
+        image: aojea/kindnetd:v1.0.0
+        env:
```
hellt, line 76: I think we should add another ENV here to drive the disablement of TX offload on that management bridge. This is needed for srlinux and other platforms that do not accept frames with fake checksums

```yaml
- name: DISABLE_CNI_BRIDGE_OFFLOAD
  value: "true"
```

deploy/deploy.go
marcushines: it can be used for anything - this lets us inject any other services

manifests/kind/kind-bridge.yaml
marcushines: added

deploy/deploy.go
alexmasi, line 278: failed to apply additional manifest

deploy/deploy.go
```diff
@@ -219,6 +219,8 @@ type KindSpec struct {
 	Kubecfg                  string            `yaml:"kubecfg"`
 	GoogleArtifactRegistries []string          `yaml:"googleArtifactRegistries"`
 	ContainerImages          map[string]string `yaml:"containerImages"`
+	KindConfigFile           string            `yaml:"config"`
+	AdditionalManifests      []string          `yaml:"additionalManifests"`
```
alexmasi, line 223: this field could technically be used to apply manifests that are not kind related, but not sure of a better place to put this field

deploy/deploy.go
```diff
@@ -262,10 +264,20 @@ func (k *KindSpec) Deploy(ctx context.Context) error {
 	if k.Kubecfg != "" {
 		args = append(args, "--kubeconfig", k.Kubecfg)
 	}
+	if k.KindConfigFile != "" {
+		args = append(args, "--config", k.KindConfigFile)
+	}
+	log.Infof("Creating kind cluster with: %s", args)
```
alexmasi, line 270: Creating kind cluster with args: %v

manifests/kind/kind-bridge.yaml
```diff
@@ -0,0 +1,123 @@
...
+        k8s-app: kindnet
+    spec:
+      hostNetwork: true
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      serviceAccountName: kindnet
+      initContainers:
+      - name: install-cni-bin
+        image: aojea/kindnetd:v1.0.0
```
alexmasi, line 68: gonna have to add this to the list of containers we rehost

Just a general question, does bridge work with all vendors, or only should be used with nokia-only topos?

alexmasi (approved):

deploy/deploy.go
marcushines: this area it is really "part of" kind deployment so i think it stands alone
if we want to make a post cluster bringup hook we could do that but for now i think this is as good a place of as any

alexmasi (approved):

manifests/kind/kind-bridge.yaml
marcushines: it works with all vendors

PR #160, opened 2022-06-30
update kindnet to include tx checksum offload

marcushines:


manifests/kind/kind-bridge.yaml
```diff
@@ -84,8 +84,6 @@ spec:
               fieldPath: status.podIP
         - name: CNI_BRIDGE
           value: "true"
-        - name: DISABLE_CNI_BRIDGE_OFFLOAD
-          value: "true"
```
hellt, line 88: what is the reason for removal?

marcushines: cause i am an idiot

guoshiuan (approved):

alexmasi (approved):

PR #162, opened 2022-07-08
Implement Configuration reset for Juniper Nodes

jinsun-yoo:


topo/node/cptx/cptx.go
```diff
@@ -208,6 +208,36 @@ func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
 	return nil
 }
 
+func (n *Node) ResetCfg(ctx context.Context) error {
+	log.Infof("%s resetting config", n.Name())
```
alexmasi, line 212: nit: lets keep this consistent with the other config related log messages

%s - resetting config

topo/node/cptx/cptx.go
```diff
@@ -208,6 +208,36 @@ func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
...
+	if err != nil {
+		return err
+	}
+
+	defer n.cliConn.Close()
+
+	cfgs := []string{
+		"load factory-default",
+		"delete system commit factory-settings",
+		"set system root-authentication encrypted-password $6$7uA5z8vs$cmHIvL0aLU4ioWAHPR0PLeU/mJj.JO/5pQVQoqRlInK3fJNTLYLhwiDi.Q6gHhltSB3S1P/.raEsuDSH7akcJ/",
```
alexmasi, line 224: I'm a little conflicted on this, on one hand an encrypted password is best practice and on the other how will a user know the default password. In our example configs this is less of a problem because they are just examples. Users do not have to use them. Here we are setting up this encrypted password as part of kne_cli itself

I also doubt in this case the security concerns of simply using the unencrypted password. Maybe a middle ground is to comment the unencrypted password on this line so our users will still know the password without it appearing directly in the config itself?

topo/node/cptx/cptx.go
```diff
@@ -208,6 +208,36 @@ func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
...
+		// SSH is necessary to connect and control the client (kubectl exec)
+		"set system services ssh root-login allow",
```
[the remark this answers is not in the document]
jinsun-yoo: Just verified that I could.

alexmasi (dismissed):

topo/node/cptx/cptx.go
jinsun-yoo: Sure, will do for arista

topo/node/cptx/cptx.go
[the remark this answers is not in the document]
jinsun-yoo: I thought I could ssh into ceos routers after resetting the config, but I can't reproduce connecting via SSH now (most likely I must have SSH-ed into another router that was not reset and had the original configs).

topo/node/cptx/cptx.go
jinsun-yoo: I will add the comments for arista on a different PR after we hear back from them.

topo/node/cptx/cptx_test.go
```diff
@@ -129,9 +130,15 @@ func TestConfigPush(t *testing.T) {
 			}
 			defer fp.Close()
 
+			/* Using scrapl_cfg_testing results in an EOF error when config includes comments.
+			 * We remove the first line of the sample config, which holds the plaintext password.
+			 * Comments in config files are not problematic when using kne_cli. */
+			fbuf := bufio.NewReader(fp)
+			_, _ = fbuf.ReadBytes('\n')
```
alexmasi, line 137: even though an err here is almost impossible, it should still be checked and t.Fatal if the err is non-nil

topo/node/cptx/cptx_test.go
```diff
@@ -129,9 +130,15 @@ func TestConfigPush(t *testing.T) {
 			}
 			defer fp.Close()
 
+			/* Using scrapl_cfg_testing results in an EOF error when config includes comments.
+			 * We remove the first line of the sample config, which holds the plaintext password.
+			 * Comments in config files are not problematic when using kne_cli. */
+			fbuf := bufio.NewReader(fp)
```
alexmasi, line 136: this is a simple solution which works for this case, my concern is that it will cause issues if the test cases changes in the future

it is safer here to write a small helper func which removes comment lines from a config and returns the comment-free version

topo/node/cptx/cptx_test.go
```diff
@@ -44,6 +48,33 @@ func (f *fakeWatch) ResultChan() <-chan watch.Event {
 	return eCh
 }
 
+// removeCommentsFromConfig removes comment lines from a JunOS config file
+// and returns the remaining config in an io.Reader.
+// Using scrapl_cfg_testing results in an EOF error when config includes comments.
+// Comments in config files are not problematic when using kne_cli (not testing).
+// This is a simple implementation that only removes lines that are entirely comments.
+func removeCommentsFromConfig(f *os.File) (io.Reader, error) {
```
alexmasi, line 56: this func is a good candidate to be a test helper: go/go-style/decisions#mark-test-helpers

then the signature can be:

func removeCommentsFromConfig(t *testing.T, io.Reader) io.Reader {}

topo/node/cptx/cptx_test.go
```diff
@@ -44,6 +48,33 @@ func (f *fakeWatch) ResultChan() <-chan watch.Event {
...
+		if err != nil && err != io.EOF {
+			return nil, fmt.Errorf("br.ReadBytes() failed: %+v\n", err)
+		}
+
+		if regexp.Find(line) != nil {
+			continue
+		}
+		fmt.Fprint(&buf, string(line))
+
+		if err == io.EOF {
```
alexmasi, line 71: would it make more sense to check this right after (or even in the initial err check)?

```
line, err := ...
if err != nil {
  if err == io.EOF {
    break
  }
  return nil, err
}
```

then

```
if re.Find(line) == nil {
  fmt.Fprint(&buf, string(line))
}
```

Those changes should reduce the number of control flow statements and make it slightly more obvious of what the code is doing. What are your thoughts?

topo/node/cptx/cptx_test.go
```diff
@@ -44,6 +48,33 @@ func (f *fakeWatch) ResultChan() <-chan watch.Event {
...
 
+// removeCommentsFromConfig removes comment lines from a JunOS config file
+// and returns the remaining config in an io.Reader.
+// Using scrapl_cfg_testing results in an EOF error when config includes comments.
+// Comments in config files are not problematic when using kne_cli (not testing).
+// This is a simple implementation that only removes lines that are entirely comments.
+func removeCommentsFromConfig(f *os.File) (io.Reader, error) {
+	var buf bytes.Buffer
+	br := bufio.NewReader(f)
+	regexp := regexp.MustCompile(`^\s*(?:(?:\/\*)|[#\*])`)
```
alexmasi, line 59: prefer naming the var `re` due to the overload on regexp

topo/node/cptx/cptx_test.go
```diff
@@ -44,6 +48,33 @@ func (f *fakeWatch) ResultChan() <-chan watch.Event {
 	return eCh
 }
 
+// removeCommentsFromConfig removes comment lines from a JunOS config file
+// and returns the remaining config in an io.Reader.
+// Using scrapl_cfg_testing results in an EOF error when config includes comments.
```
alexmasi, line 53: scrapli

topo/node/cptx/cptx_test.go
jinsun-yoo: Looking at this once more, I realized I can still use the change below regardless of how we handle the error:

```
if re.Find(line) == nil {
  fmt.Fprint(&buf, string(line))
}
```

alexmasi: Oh I wasn't aware of that behavior, good with me then

alexmasi (approved):

PR #164, opened 2022-07-13
Add custom privilege level patterns in Juniper to allow hostnames with '.'

jinsun-yoo:


topo/node/cptx/cptx_test.go
```diff
@@ -139,6 +140,44 @@ func TestConfigPush(t *testing.T) {
...
+
+	privilegePromptMap := map[string]string{
+		"exec":          "root@%s>",
+		"configuration": "root@%s#",
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			for privilege, prompt := range privilegePromptMap {
+				match, err := regexp.Match(privLevels[privilege].Pattern, []byte(fmt.Sprintf(prompt, tt.hostname)))
```
alexmasi, line 169: Safer to lookup the dict first in case key DNE:

```
fullPrompt := fmt.Sprintf(prompt, tt.hostname)
privLevel, ok := privLevels[privilege]
if !ok {
  t.Fatalf(...)
}
match, err := regexp.MatchString(privLevel.Pattern, fullPrompt)
if err != nil {
  t.Fatalf("regexp.MatchString() failed: %v", err)
}
if match != tt.shouldMatch {
  t.Fatal("regexp.MatchString() unexpected result (got %v, want %v) for prompt %q and privilege level %q", match, tt.shouldMatch, fullPrompt, privLevel)
}
```

Also note a few more patterns:
- error message should contain the func that failed
- got before want
- error message is clear exactly what case failed to the user (without even needing to lookup the test case by description name)

topo/node/cptx/cptx_test.go
```diff
@@ -139,6 +140,44 @@ func TestConfigPush(t *testing.T) {
...
+			shouldMatch: true,
+		},
+		{
+			desc:        "hostname with '.'",
+			hostname:    "test.example",
+			shouldMatch: true,
+		},
+	}
+
+	privilegePromptMap := map[string]string{
```
alexmasi, line 161: nit: usually the pattern for tests that we follow is:

things needed for all test cases

testcase def

run tests

So in this case
```
privilegePromptMap := ...

tests := []struct...
for _, tt := range tests ...
```

topo/node/cptx/cptx_test.go
```diff
@@ -139,6 +140,54 @@ func TestConfigPush(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			for privilegeName, prompt := range privilegePromptMap {
+				fullPrompt := fmt.Sprintf(prompt, tt.hostname)
+				privLevel, ok := customPrivLevels[privilegeName]
+				if !ok {
+					t.Fatalf("privilege %q not defined in custom privilege", privilegeName)
+				}
+				match, err := regexp.Match(privLevel.Pattern, []byte(fullPrompt))
```
alexmasi, line 179: did you forget to change it to regexp.MatchString ? Or does that not work

Otherwise LGTM

alexmasi (approved):

topo/node/cptx/cptx_test.go
jinsun-yoo: @alexmasi  I simply forgot to change this. Could you give me one more LGTM?

alexmasi (approved):

PR #166, opened 2022-07-22
`kind-bridge` as a default kind cluster config

hellt:
Hi all,
in this PR I propose to set kind-bridge as a default kne+kind cluster config as this allows us to make sure users will be able to use all the images kne has support for, without requiring them be aware of limitations default PTP plugin imposes.

Additionally, in this PR 

* make sure that kind config file path also follows the same logic of using relative paths.
* added a make target to invoke super linter


marcushines (dismissed):
just need to fix lint

hellt:
@marcushines I tried to squash linter errors, but for whatever reason textlint doesn't eat my rc file and complaints about `repo` and `node` words.
Maybe someone who added that super-linter can give a hand here... Other than that, the rest looks clean

hellt:
okay, found my way around text linter
now all is green, much better

marcushines (approved):

PR #167, opened 2022-07-22
Add srlinux controller deployment to `kne_cli deploy`

alexmasi:


alexmasi:
After the latest commit the linter (`golangci-lint run -c .github/linters/.golangci.yml ./...`) runs with no issues locally. However the github action linter is reporting issues with code spread across multiple files in the same package coming from the `typecheck` linter which is not enabled in our config file.

hellt:
@alexmasi I think this is because super-linter is configured to run on the partial repo (due to the env var set). This makes go ci linter to see only changed files, hence failing the test.

Maybe it makes sense to try run the CI on the whole repo?

controller/server/main.go
```diff
@@ -43,6 +43,7 @@ var (
 	defaultMetallbManifestDir = ""
 	defaultMeshnetManifestDir = ""
 	defaultIxiaTGManifestDir  = ""
+	defaultSRLinuxManifestDir = ""
```
hellt, line 46: does it make sense to default to srl-controller https uri path as a default for this var?

hellt: looks like controller/server is something used internally, hence the paths are relative to homedir

alexmasi: Yep that's the case

PR #168, opened 2022-07-23
`kne` instead of `kne_cli`

hellt:
Hi @marcushines @alexmasi 
This proposal is to retire usage of `kne_cli` as a binary name of a kne CLI in favor of just `kne`.

From my experience, and what seems to be in line with opinionated cli guidelines, a good CLI name must be short, consice and easy to type -  https://clig.dev/#naming.

`kne` in that regard is better than `kne_cli` as it is 4 chars shorter, doesn't force you to use SHIFT and is as expressive in what it does.

I know this might be a bit too intrusive, but I am afraid if we don't do this early on, it might be harder to change it in the future.

I will leave a few inline comments over the certain code lines which I feel deserve a bit of explanatoin


.mk/ocipush.mk
```diff
@@ -0,0 +1,14 @@
...
+oci-push: build
+	@echo
+	
+	@echo "With the following pull command you get a kne binary at your working directory. To use this downloaded binary use - './kne <command>'\nMake sure to add ./ prefix in order to use the downloaded binary and not the globally installed kne"
+	
+	@echo 'If https proxy is configured in your environment, pass the proxies via --env HTTPS_PROXY="<proxy-address>" flag of the docker run command.'
+	
+# push to ttl.sh
+	@docker run --rm -v $$(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 push ttl.sh/kne-$(COMMIT):1d ./kne > /dev/null
+	@echo "download with: docker run --rm -v \$$(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ttl.sh/kne-$(COMMIT):1d"
```
hellt, line 14: this new target `oci-push` is a great way to share built binaries with anybody on the internet. No need to ask users to clone/build the project if all you want is for them to test the code off of your working branch.

This target builds the `kne` bin and uploads it to anonymous container registry - ttl.sh (https://ttl.sh)

This registry stores the artifact for up to max 1d, doesn't require login, and makes it super easy to exchange bins. The receiving side just runs a single docker pull command and get the binary in their working dir immediately.

Here is how it works. The kne dev calls a target:

```
❯ make oci-push 
CGO_ENABLED=0 go build -o kne -ldflags="-s -w" kne_cli/main.go

With the following pull command you get a kne binary at your working directory. To use this downloaded binary use - './kne <command>'
Make sure to add ./ prefix in order to use the downloaded binary and not the globally installed kne
If https proxy is configured in your environment, pass the proxies via --env HTTPS_PROXY="<proxy-address>" flag of the docker run command.
download with: docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ttl.sh/kne-adeeba3:1d
```

Then they share the docker run command (or the whole output with some explanations given in it) with the receiving party. The receiver does:

```
❯ docker run --rm -v $(pwd):/workspace ghcr.io/oras-project/oras:v0.12.0 pull ttl.sh/kne-adeeba3:1d
Downloaded 5334cb06a3da kne
Pulled ttl.sh/kne-adeeba3:1d
Digest: sha256:d8ea741fbf9f50fba0373282dd0e4524e1a3d3a55d0a53428d7250c26c79c969
```
and just as that, they have kne binary in their PWD
```
❯ file kne 
kne: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=VY5rDjJWfdO4BHB68L_W/nVfe6_DAaO0_kmrRWxn8/LVEkBjXxLG3NHqZ2XPC-/5H1QvHQT79qz8JuE-jZA, stripped
```

Now, in containerlab, since I have maintainer rights, I also push the binaries to ghcr.io registry, which is a bit faster to push and pull from. I don't have contributors rights for KNE repo, so left this part out of the PR, but later on we can do that.

Makefile
```diff
@@ -24,3 +27,13 @@ up: kind-start
...
+
+.PHONY: build
+## Build kne
+build:
+	CGO_ENABLED=0 go build -o $(KNE_CLI_BIN) -ldflags="-s -w" kne_cli/main.go
+
+.PHONY: install
+## Install kne cli binary to user's local bin dir
+install: build
+	mv $(KNE_CLI_BIN) $(INSTALL_DIR)
```
hellt, line 39: install target that I propose to use instead of `go install` variant.

cloudbuild/external.pkr.hcl
```diff
@@ -104,8 +104,8 @@ build {
       "/home/$USER/go/bin/go-licenses check github.com/openconfig/kne/kne_cli",
       "/home/$USER/go/bin/go-licenses save github.com/openconfig/kne/kne_cli --save_path=\"../third_party/licenses/kne_cli\"",
       "cd kne_cli",
-      "/usr/local/go/bin/go build -v",
-      "sudo cp kne_cli /usr/local/bin/",
+      "/usr/local/go/bin/go build -v -o kne",
+      "sudo cp kne /usr/local/bin/",
```
hellt, line 108: the packer stuff I have no way to test, so it would be cool if you give it a go...

marcushines (approved):

marcushines (approved):

PR #182, opened 2022-07-30
implementing resetter for srlinux

Hobbydos:


hellt:
if we need tests for that, I can have a look at it

alexmasi:
Yes, a unit test for ResetCfg implementation would be good. Here is an example of a unit test for cEOS implementation: https://github.com/openconfig/kne/blob/737516aac0a5477770eb2b96dd8f4c05de8a793b/topo/node/ceos/ceos_test.go#L303 (which looks like you wrote a good chunk of)

marcushines (approved):

marcushines (commented):
please add a test

hellt:
@marcushines unit tests are added in #187 which should supersede this PR

hellt:
With #187 merfed this can be safely closed

PR #183, opened 2022-08-01
Support push + reset cfg from topology manager grpc service

alexmasi:


controller/server/main.go
```diff
@@ -356,6 +357,60 @@ func (s *server) ShowTopology(ctx context.Context, req *cpb.ShowTopologyRequest)
 	return resp, nil
 }
 
+func (s *server) PushConfig(ctx context.Context, req *cpb.PushConfigRequest) (*cpb.PushConfigResponse, error) {
+	log.Infof("Received PushConfig request: %v", req)
+	s.muTopo.Lock()
```
marcushines, line 362: do we really want a  global lock for this operation?

marcushines (approved):

topo/topo_test.go
```diff
@@ -184,6 +184,10 @@ func (f *defaultFakeTopology) ConfigPush(context.Context, string, io.Reader) err
 	return nil
```
marcushines, line 184: why no test?

controller/server/main.go
```diff
@@ -356,6 +357,60 @@ func (s *server) ShowTopology(ctx context.Context, req *cpb.ShowTopologyRequest)
 	return resp, nil
 }
 
+func (s *server) PushConfig(ctx context.Context, req *cpb.PushConfigRequest) (*cpb.PushConfigResponse, error) {
```
marcushines, line 360: why no tests?

controller/server/main.go
alexmasi: I decided to use the lock here to prevent a request to a different method from changing a topo while this method relies on that topos state and existence.

This would be more efficient if we had one lock per topology (for edits like these) and only use this lock for updating the topology map

topo/topo_test.go
alexmasi: yeah looks like this file in general needs more coverage for all of the topo manager methods

I added a task to increase the coverage for this file in general

controller/server/main.go
alexmasi: right now there is a test for this binary that only tests that the incoming request can be parsed into the correct struct for a deployment

there should be coverage for all of the grpc method implementations (not only the two new ones added), most of them are basically just calling the topology manager library

topo/topo.go
```diff
@@ -117,31 +95,29 @@ func WithClusterConfig(r *rest.Config) Option {
...
 func WithBasePath(s string) Option {
 	return func(m *Manager) {
-		m.BasePath = s
+		m.basePath = s
 	}
 }
 
-// New creates a new topology manager based on the provided kubecfg and topology.
-func New(kubecfg string, pb *tpb.Topology, opts ...Option) (TopologyManager, error) {
+// New creates a new Manager based on the provided topology.
```
marcushines, line 102: add to the doc string how rcfg > clustercfg > kubecfg

topo/topo.go
```diff
@@ -162,26 +138,124 @@ func New(kubecfg string, pb *tpb.Topology, opts ...Option) (TopologyManager, err
...
+	sort.Strings(keys)
+	var n []node.Node
+	for _, k := range keys {
+		n = append(n, m.nodes[k])
+	}
+	return n
+}
+
+// load populates the internal fields of the topology proto.
+func (m *Manager) load(ctx context.Context) error {
```
marcushines, line 244: remove unneeded param?

topo/topo.go
```diff
@@ -677,92 +597,62 @@ func serviceToProto(s *corev1.Service, m map[uint32]*tpb.Service) error {
...
+	b, err := os.ReadFile(path)
 	if err != nil {
 		return nil, err
 	}
-	for _, n := range t.TopologyProto().Nodes {
-		if len(n.Services) == 0 {
-			n.Services = map[uint32]*tpb.Service{}
+	t := &tpb.Topology{}
+	switch {
+	case strings.HasSuffix(path, ".yaml") || strings.HasSuffix(path, ".yml"):
```
marcushines, line 644: case strings.HasSuffix(path, ".yaml"), strings.HasSuffix(path, ".yml"):

cmd/topology/topology.go
```diff
@@ -98,38 +98,34 @@ func resetCfgFn(cmd *cobra.Command, args []string) error {
...
-	t.Load(ctx)
-	nodes := t.Nodes()
-	if len(args) > 1 {
-		n, err := t.Node(args[1])
-		if err != nil {
-			return err
-		}
-		nodes = []node.Node{n}
+	nodes := map[string]node.Node{}
+	for _, n := range tm.Nodes() {
```
marcushines, line 107: hmm why not have Nodes() return map?

cmd/root.go
```diff
@@ -40,7 +40,7 @@ var (
 	rootCmd = &cobra.Command{
 		Use:   "kne",
 		Short: "Kubernetes Network Emulation CLI",
-		Long: `Kubernetes Network Emulation CLI.  Works with meshnet to create 
+		Long: `Kubernetes Network Emulation CLI.  Works with meshnet to create
```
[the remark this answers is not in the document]
alexmasi: The line breaks are kept so the space is not needed:

```
$ ./kne_cli --help
Kubernetes Network Emulation CLI.  Works with meshnet to create
layer 2 topology used by containers to layout networks in a k8s
environment.

Usage:
  kne [command]
...
```

cmd/topology/topology.go
alexmasi: done

api/clientset/v1beta1/topology.go
```diff
@@ -129,6 +129,10 @@ func (t *topologyClient) Get(ctx context.Context, name string, opts metav1.GetOp
 }
 
 func (t *topologyClient) Create(ctx context.Context, topology *topologyv1.Topology, opts metav1.CreateOptions) (*topologyv1.Topology, error) {
+	topology.TypeMeta = metav1.TypeMeta{
+		Kind:       "Topology",
+		APIVersion: "networkop.co.uk/v1beta1",
+	}
```
marcushines, line 135: with this create does that fix one of the tests which that never got set?

marcushines (dismissed):

api/clientset/v1beta1/topology.go
alexmasi: oh you're right I forgot about that, let me check

alexmasi: yup, fixed the test

marcushines (approved):

PR #185, opened 2022-08-03
Add proper Create signature

marcushines:
Add coverage for Update

### DIFF
--- a/api/clientset/v1beta1/topology_test.go
+++ b/api/clientset/v1beta1/topology_test.go
@@ -14,12 +14,14 @@ import (
 	"github.com/kr/pretty"
 	topologyv1 "github.com/openconfig/kne/api/types/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	dynamicfake "k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	restfake "k8s.io/client-go/rest/fake"
+	ktest "k8s.io/client-go/testing"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


api/clientset/v1beta1/topology_test.go
```diff
@@ -78,7 +80,26 @@ func setUp(t *testing.T) (*Clientset, *restfake.RESTClient) {
...
 	cs.restClient = fakeClient
-	cs.dInterface = dynamicfake.NewSimpleDynamicClient(scheme.Scheme, objs...).Resource(gvr)
+	f := dynamicfake.NewSimpleDynamicClient(scheme.Scheme, objs...)
+	// Add handler for Update call.
+	f.PrependReactor("update", "*", func(action ktest.Action) (bool, runtime.Object, error) {
+		uAction, ok := action.(ktest.UpdateAction)
+		if !ok {
+			return false, nil, nil
+		}
+		fmt.Println(action)
```
alexmasi, line 90: Add context to what is being printed for looking at the test logs

(here and with tObj)

api/clientset/v1beta1/topology_test.go
```diff
@@ -111,7 +132,7 @@ func TestCreate(t *testing.T) {
 		}
 		t.Run(tt.desc, func(t *testing.T) {
 			tc := cs.Topology("foo")
-			got, err := tc.Create(context.Background(), tt.want)
+			got, err := tc.Create(context.Background(), tt.want, metav1.CreateOptions{})
```
alexmasi, line 135: Would a test case that sets one of these options be valuable here?

api/clientset/v1beta1/topology_test.go
```diff
@@ -289,11 +310,59 @@ func TestWatch(t *testing.T) {
...
+				t.Fatalf("failed to generate update: %v", err)
+			}
+			got, err := tc.Update(context.Background(), &unstructured.Unstructured{Object: update}, metav1.UpdateOptions{})
+			if s := errdiff.Substring(err, tt.wantErr); s != "" {
+				t.Fatalf("unexpected error: %s", s)
+			}
+			if tt.wantErr != "" {
+				return
+			}
+			want := tt.want.DeepCopy()
```
alexmasi, line 353: why do you need a deep copy and comparison with reflection here instead of cmp/opts?

alexmasi (approved):

api/clientset/v1beta1/topology_test.go
marcushines: nope you never set these in any general case

api/clientset/v1beta1/topology_test.go
marcushines: cause cmp wasn't widely used when first did this i will flip it at some point

alexmasi (approved):

api/clientset/v1beta1/topology_test.go
hellt: if you won't do a deepcopy, won't you change the `obj` vars in-place as is is a pointer, and risk mutating an object

PR #192, opened 2022-08-18
Fix handling for KUBECONFIG env var

marcushines:


cmd/root_test.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"os"
+	"testing"
+)
+
+func TestGetKubeCfg(t *testing.T) {
+	tests := []struct {
+		desc   string
+		setVar func() func()
+		osVar  string
```
alexmasi, line 12: remove unused

cmd/root.go
```diff
@@ -62,12 +61,19 @@ func ExecuteContext(ctx context.Context) error {
 	return rootCmd.ExecuteContext(ctx)
 }
 
-func init() {
+func getKubeCfg() string {
```
alexmasi, line 64: I know this codebase is not held to the same internal style guide, but worth mentioning: go/go-style/decisions#getters

alexmasi (dismissed):
Can you add the logging of which kubecfg is used to this PR as well?

alexmasi (approved):

PR #195, opened 2022-08-29
KNE to use the arista-ceoslab-operator

frasieroh:
These changes are to integrate the new cEOS-lab operator (shared privately with Marcus and Alex) with KNE. My assumption is that you guys would take it for a spin and if it looks alright we can proceed with open-sourcing and merging this PR. We have a small amount of red tape to cut through on our end.

This only works with a very recent version of cEOS because of a required fix to how read-only boot scripts are handled at startup. I'll be sharing this with you shortly and integrating this into the next drop.


topo/node/ceos/ceos.go
```diff
@@ -49,9 +52,16 @@ type Node struct {
 
 // Add validations for interfaces the node provides
 var (
-	_ node.Certer       = (*Node)(nil)
```
alexmasi, line 52: Please keep the Certer implementation but have it return an [unimplemented error](https://pkg.go.dev/google.golang.org/grpc@v1.48.0/codes#pkg-types) with an adequate error message stating exactly how a user should create the certs

so if a user tries to create new self signed certs for the node using the kne_cli they can be pointed to the right way to do it (send a request to the controller / certs must be included in the topo textproto at startup... etc.)

topo/node/ceos/ceos.go
```diff
@@ -66,99 +76,143 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
...
-		log.Infof("%s - no cert config", n.Name())
-		return nil
+func (n *Node) CreateCRD(ctx context.Context) error {
+	log.Infof("Creating new CEosLabDevice CRD for node: %v", n.Name())
+	proto := n.GetProto()
+	config := proto.GetConfig()
+	device := &ceos.CEosLabDevice{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "ceoslab.arista.com/v1alpha1",
+			Kind:       "CEosLabDevice",
```
alexmasi, line 103: just curious why its capitalized like this?

topo/node/ceos/ceos.go
```diff
@@ -66,99 +76,143 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
...
-		).String(),
+		FieldSelector: fields.SelectorFromSet(fields.Set{metav1.ObjectNameField: n.Name()}).String(),
 	})
 	if err != nil {
 		return err
 	}
 	for e := range w.ResultChan() {
 		p := e.Object.(*corev1.Pod)
-		if p.Status.Phase == corev1.PodRunning {
+		if p.Status.Phase == corev1.PodPending {
```
alexmasi, line 171: Why is the behavior changed to wait for Pending instead of Running?

topo/node/ceos/ceos.go
```diff
@@ -329,15 +340,21 @@ func defaults(pb *tpb.Node) *tpb.Node {
 	return pb
 }
 
-func (n *Node) FixInterfaces() {
+func (n *Node) FixInterfaces() error {
 	for k, v := range n.Proto.Interfaces {
 		if !strings.HasPrefix(k, "eth") {
```
alexmasi, line 345: a switch statement would be more clear here:

switch {
default:
  return fmt.Errorf("unrecognized interface name: %s", v.Name)
case !strings.HasPrefix(k, "eth"), ethIntfRe.MatchString(v.Name), mgmtIntfRe.MatchString(v.Name):
case v.Name == "":
  n.Proto.Interfaces[k].Name = fmt.Sprintf("Ethernet%s", strings.TrimPrefix(k, "eth"))
}
return nil

topo/node/ceos/ceos_test.go
```diff
@@ -66,34 +66,67 @@ func TestNew(t *testing.T) {
...
-						"/sbin/init",
-						"systemd.setenv=INTFTYPE=eth",
-						"systemd.setenv=ETBA=1",
-						"systemd.setenv=SKIP_ZEROTOUCH_BARRIER_IN_SYSDBINIT=1",
-						"systemd.setenv=CEOS=1",
-						"systemd.setenv=EOS_PLATFORM=ceoslab",
-						"systemd.setenv=container=docker",
-					},
+					// The operator handles defaults for command, env, and image, args
+					Command:      nil,
```
alexmasi, line 122: empty fields can be omitted

topo/node/ceos/ceos_test.go
```diff
@@ -200,106 +250,6 @@ func TestNew(t *testing.T) {
 	}
 }
 
-func TestGenerateSelfSigned(t *testing.T) {
```
alexmasi, line 203: are there tests in the controller for the generation of self signed certs?

topo/node/ceos/ceos.go
```diff
@@ -248,10 +302,10 @@ func defaults(pb *tpb.Node) *tpb.Node {
 	}
 	if pb.Labels == nil {
```
alexmasi, line 303: this can be simplified:

if pb.Labels == nil {
  pb.Labels = map[string]string{}
}
if pb.Labels["type"] == "" {
  pb.Labels["type"] = tpb.Node_ARISTA_CEOS.String()
}
if pb.Labels["vendor"] == "" {
  pb.Labels["vendor"] = tpb.Vendor_ARISTA.String()
}
if pb.Labels["model"] == "" {
  pb.Labels["model"] = pb.Model
}
if pb.Labels["os"] == "" {
  pb.Labels["os"] = pb.Os
}
if pb.Labels["version"] == "" {
  pb.Labels["version"] = pb.Version
}

topo/node/ceos/ceos.go
```diff
@@ -66,99 +76,143 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
...
+	}
+	for k, v := range n.GetProto().GetInterfaces() {
+		if device.Spec.IntfMapping == nil {
+			device.Spec.IntfMapping = map[string]string{}
+		}
+		device.Spec.IntfMapping[k] = v.GetName()
+	}
+	// Post to k8s
+	r := ceos.CEosLabDevice{}
+	err := n.KubeClient.CoreV1().RESTClient().
```
alexmasi, line 154: check this error

topo/node/ceos/ceos.go
```diff
@@ -49,9 +52,16 @@ type Node struct {
...
-	_ node.Certer       = (*Node)(nil)
 	_ node.ConfigPusher = (*Node)(nil)
 	_ node.Resetter     = (*Node)(nil)
+
+	ethIntfRe  = regexp.MustCompile(`^Ethernet\d+(?:/\d+)?(?:/\d+)?$`)
+	mgmtIntfRe = regexp.MustCompile(`^Management\d+(?:/\d+)?$`)
+)
+
+const (
+	API      = "ceoslab.arista.com/v1alpha1"
```
alexmasi, line 63: these constants should not be CAPS

this also has the side effect of exporting them

all lower case is preferred here

topo/node/ceos/ceos.go
frasieroh: The correct marketing format is `cEOS-lab`, but I wanted to abide by Kubernetes-isms, which seem to be to be PascalCase for kinds.

topo/node/ceos/ceos.go
frasieroh: Thanks for catching this. I install a startup probe in the operator and I was experimenting with blocking on that, but it doesn't appear the core API exports that state as directly (would have had to iterate over pod conditions) and I decided not to move forward with it. It would also have had the side effect of not allowing pods to "boot" concurrently. I must have botched it when going back.

topo/node/ceos/ceos.go
frasieroh: +1

topo/node/ceos/ceos_test.go
frasieroh: +1

topo/node/ceos/ceos.go
frasieroh: +1

topo/node/ceos/ceos.go
frasieroh: +1

topo/node/ceos/ceos_test.go
frasieroh: I manually validated that it worked and didn't dwell on it (the device does it's own sanity testing when certs are loaded) but I can write some and link the commit back here.

frasieroh: Actually, there isn't much I can do to validate the contents of the certificates given the body is effectively random. [This test](https://github.com/aristanetworks/arista-ceoslab-operator/blob/main/test/reconcile-certs/01-assert.yaml) already validates:

- Certificates are configured as secrets
- They are defined as volumes in the pod
- The pod is running

Do to some deficiencies in kuttl I can't validate the volumeMounts are correct: https://github.com/kudobuilder/kuttl/issues/76

The main purpose of this test seems to be to validate KNE's use of the EOS CLI. This isn't relevant from the controller since we're generating them with the go standard `x509` lib. Errors in that are just checked.

The only way a user could mess this up, I suppose, would be to give multiple certs the same name? There isn't much I can really do about that. K8s would report the conflicting mount when creating the pod, and this would show up in the controller's logs. A better approach would be to sanitize the topology (you might already do this, didn't check).

alexmasi: Okay I just wanted to make sure we weren't losing coverage here from an overall perspective, sgtm

frasieroh: Woops, switch volumes with volumeMounts and vice-versa.

topo/node/ceos/ceos.go
```diff
@@ -66,42 +79,95 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
...
+			Name:      n.Name(),
+			Namespace: n.GetNamespace(),
+			Labels: map[string]string{
+				"app":  n.Name(),
+				"topo": n.GetNamespace(),
+			},
+		},
+		Spec: ceos.CEosLabDeviceSpec{
+			EnvVar:        config.GetEnv(),
+			Image:         config.GetImage(),
```
alexmasi, line 118: we also need the init image field to be configurable

topo/node/ceos/ceos.go
frasieroh: As we discussed not sure how I got so confused about this, but this was originally PodPending. Setting it to PodRunning introduced a deadlock whereby init containers would be waiting for pods whose creation was blocked here.

topo/node/ceos/ceos.go
frasieroh: Pushed arista-ceoslab-operator v0.2.0 and updated PR

alexmasi:
Okay things look good on my end, once the controller repo + controller container are opensourced and a mainline version of ceos that contains the necessary fixes for cert gen is released and ingested by us we can get this merged

topo/node/ceos/ceos.go
```diff
@@ -66,99 +79,154 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
...
+		if device.Spec.IntfMapping == nil {
+			device.Spec.IntfMapping = map[string]string{}
+		}
+		device.Spec.IntfMapping[k] = v.GetName()
 	}
-	log.Infof("%s - generating self signed certs", n.Name())
-	log.Infof("%s - waiting for pod to be running", n.Name())
+	// Post to k8s
+	r := ceos.CEosLabDevice{}
+	err := n.KubeClient.CoreV1().RESTClient().
```
alexmasi, line 158: sorry I missed this in the first review, can you abstract these details into a clientset and then call client.Create here (same with Delete)? An example is our CRD for topology: https://github.com/openconfig/kne/blob/b09a14a6ff36eb0a9448b5b95bac5d5ec9611da1/api/clientset/v1beta1/topology.go#L132

Here is an example a few months ago when we also had a rest client: https://github.com/openconfig/kne/blob/a2a782b79881cc7735cd11214313ebf45e6ea779/api/clientset/v1beta1/topology.go#L148

This will make testing a lot easier using fake clientsets

frasieroh: okay, I defined a clientset in the operator. I may get to a fake clientset later but we don't need one for this PR. k8s code generators work poorly with operator-sdk.

topo/node/ceos/ceos.go
```diff
@@ -27,9 +28,14 @@ import (
...
 	scrapliutil "github.com/scrapli/scrapligo/util"
 	log "github.com/sirupsen/logrus"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+
+	ceos "github.com/aristanetworks/arista-ceoslab-operator/api/v1alpha1"
+	ceosClient "github.com/aristanetworks/arista-ceoslab-operator/api/v1alpha1/clientset"
```
alexmasi, line 38: import renaming should follow package naming style and not use camelCase, prefer ceosclient or ceosv1client here

nit: also admittedly I do not know the reason why most k8 packages end in v1 or similar (https://github.com/openconfig/kne/blob/d2318708297cd379719c4f44bc73911cc3b25163/api/clientset/v1beta1/topology.go#L21) instead of the package name "clientset" for example, but we have chosen to follow that pattern with our topology CRD (https://github.com/openconfig/kne/blob/d2318708297cd379719c4f44bc73911cc3b25163/api/clientset/v1beta1/fake/fake.go#L17). up to you if you think it is worth changing your package name to match that style

topo/node/ceos/ceos.go
alexmasi: Thanks for adding that.

Btw, here is what our fake for our topology clientset looked like before we switched completely from the rest client to the dynamic client (which made testing easier): https://github.com/openconfig/kne/blob/3ad72bc0701eaaa7236f04bfe81d6a3495a93161/api/clientset/v1beta1/fake/fake.go (if that helps)

It would be nice at some point to have unit tests where a ceos is attempted to be created when one already exists, or attempted to be deleted when it DNE at some point when the fake is created.

alexmasi:
Okay once that comment is addressed I will do my end to end test again (with the newly released ceos image) to make sure everything is working one last time and then we can move forward with finally getting this merged

topo/node/ceos/ceos.go
frasieroh: ok, switched to package case

alexmasi:
okay doing a final check

alexmasi (approved):

PR #199, opened 2022-09-02
Add support kind version also adds significant coverage to deploy

marcushines:


deploy/deploy.go
```diff
@@ -222,6 +223,56 @@ type KindSpec struct {
 	AdditionalManifests      []string          `yaml:"additionalManifests"`
 }
 
+var (
+	vOut                  = bytes.NewBuffer([]byte{})
+	vExec execerInterface = kexec.NewExecer(vOut, vOut)
+)
+
+type version struct {
+	Major int
```
alexmasi, line 232: Do these need to be exported?

deploy/deploy.go
```diff
@@ -233,7 +284,36 @@ func (k *KindSpec) checkDependencies() error {
 			errs.Add(fmt.Errorf("install dependency %q to deploy", bin))
 		}
 	}
-	return errs.Err()
+	if errs.Err() != nil {
+		return errs.Err()
+	}
+	if k.Version != "" {
```
alexmasi, line 290: Where does Version even come into play? Currently its a field in the yaml but I can't find where it is used. 

If the user does not provide a version what should happen?

deploy/deploy.go
```diff
@@ -233,7 +284,36 @@ func (k *KindSpec) checkDependencies() error {
...
+		}
+
+		gotV, err := k.getKindVersion(vKindFields[1])
+		if err != nil {
+			return fmt.Errorf("kind version check failed: %w", err)
+		}
+		if gotV.Less(wantV) {
+			return fmt.Errorf("kind version check failed: got %s, want %s", gotV, wantV)
+		}
+		log.Infof("kind version valid: got %s want %s", gotV, wantV)
```
alexmasi, line 314: above (line 112) we check the existence of docker and kubectl, we should also check those versions (and likely need to check against the Image field)

deploy/deploy.go
```diff
@@ -233,7 +284,36 @@ func (k *KindSpec) checkDependencies() error {
...
+		if len(vKindFields) < 2 {
+			return fmt.Errorf("failed to parse kind version from: %s", vOut)
+		}
+
+		gotV, err := k.getKindVersion(vKindFields[1])
+		if err != nil {
+			return fmt.Errorf("kind version check failed: %w", err)
+		}
+		if gotV.Less(wantV) {
+			return fmt.Errorf("kind version check failed: got %s, want %s", gotV, wantV)
```
alexmasi, line 312: so the only binary that we would upgrade automatically would be kubectl in the Kind case?

deploy/deploy.go
marcushines: it is used right below?

marcushines: if none set - do nothing assume the user knows what they are doing

deploy/deploy.go
marcushines: this is just step one

deploy/deploy.go
marcushines: others will come later 
each should have an option to leave them or to upgrade them if they are older

deploy/deploy.go
marcushines: i am just signally to users that those 3 fields are usable

alexmasi (approved):

PR #200, opened 2022-09-06
Update metallb and refactor deploy as step two to add check for kubectl

marcushines:
version


deploy/deploy.go
```diff
@@ -134,6 +135,35 @@ func (d *Deployment) Deploy(ctx context.Context, kubecfg string) error {
...
+		return err
+	}
+	kServerVersion, err := getVersion(kubeYAML.ServerVersion.GitVersion)
+	if err != nil {
+		return err
+	}
+	origMajor := kClientVersion.Major
+	kClientVersion.Major -= 2
+	if kServerVersion.Less(kClientVersion) {
+		log.Warn("Kube client and server versions are not within expected range.")
```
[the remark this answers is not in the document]
marcushines: again i am not solving everything at once this a step - since before this didn't give users any hint now it is does

deploy/deploy.go
```diff
@@ -134,6 +135,35 @@ func (d *Deployment) Deploy(ctx context.Context, kubecfg string) error {
 	if err != nil {
 		return err
 	}
+
+	log.Infof("Checking kubectl versions.")
```
alexmasi, line 139: this block of code should have test coverage

ideally we should add a test for all of deployment.Deploy with fakeCluster/Ingress/CNI/Controller implementations

but for now we could just move this added block to its own function and write a test for that

deploy/deploy_test.go
```diff
@@ -484,11 +484,11 @@ func TestMetalLBSpec(t *testing.T) {
 			IPCount: 20,
 		},
 		execer: exec.NewFakeExecer(nil, nil, nil),
-		wantCM: `address-pools:
-    - name: default
-      protocol: layer2
-      addresses:
-        - 172.18.0.50 - 172.18.0.70
+		wantCM: `AddressPools:
```
alexmasi, line 487: with this change we should delete https://github.com/openconfig/kne/blob/main/examples/metallb/config_map.yaml (since its no longer very useful and now OOD)

alexmasi (approved):

deploy/deploy.go
marcushines: ack there needs to be a test deploy but since there isn't ... not really worth trying to address this in this pr

deploy/deploy_test.go
marcushines: will do in next pr didn't even know that was still a thing

PR #205, opened 2022-09-09
Fix port numbers on outside ports when they are remapped add test coverage for this case

marcushines:


topo/topo.go
```diff
@@ -583,7 +583,8 @@ var populateServiceMap = func(s *corev1.Service, m map[uint32]*tpb.Service) erro
 		if service.Name == "" {
 			service.Name = p.Name
 		}
-		service.Outside = uint32(p.TargetPort.IntVal)
+		service.Outside = k
+		service.Inside = uint32(p.TargetPort.IntVal)
```
guoshiuan, line 587: nit: should we remove line 579 since it will be always overwritten?

marcushines: done

guoshiuan (approved):

PR #208, opened 2022-09-12
Fix deploy to use the metallb crd

marcushines:


api/metallb/clientset/v1beta1/client_test.go
```diff
@@ -0,0 +1,399 @@
...
+			APIVersion: "metallb.io/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:       "newObj",
+			Namespace:  "test",
+			Generation: 1,
+		},
+		Status: metallbv1.IPAddressPoolStatus{},
+		Spec: metallbv1.IPAddressPoolSpec{
+			Addresses: []string{"192.168.1.100 - 192.168.1.200 "},
```
alexmasi, line 47: does this require whitespace?

deploy/deploy.go
```diff
@@ -165,6 +169,8 @@ func (d *Deployment) Deploy(ctx context.Context, kubecfg string) error {
 	vOut.Reset()
 
 	d.Ingress.SetKClient(kClient)
+	d.Ingress.SetRCfg(rCfg)
```
alexmasi, line 172: is there a downside to setting the mClient directly instead of rCfg? I think that would be more in line with how we set kClient, but maybe I'm missing a subtle difference

alexmasi (approved):

PR #209, opened 2022-09-13
fix deploy to add wait for healthy and retry based on connect error for validator

marcushines:


deploy/deploy.go
```diff
@@ -632,6 +632,12 @@ func (m *MetalLBSpec) Deploy(ctx context.Context) error {
...
 		}
 	}
+
+	// Wait for metallb to be healthy
+	if err := m.Healthy(ctx); err != nil {
+		return err
+	}
+
 	_, err = m.mClient.IPAddressPool("metallb-system").Get(ctx, "kne-service-pool", metav1.GetOptions{})
 	if err != nil {
```
alexmasi, line 642: can just check err == nil return nil here to reduce the indentation

alexmasi (approved):

PR #221, opened 2022-09-27
Add controllers to manifests

alexmasi:
The only issue here is that the user will still need to either (if they want to use ixia-c):

- manually craft and apply a config map with the desired image versions (will need to add documentation to make it extremely clear as I expect it will be encountered often)
- manually edit the deployment yaml before deploying to outline the desired images


deploy/deploy.go
```diff
@@ -827,7 +825,8 @@ func (i *IxiaTGSpec) Deploy(ctx context.Context) error {
 	if i.ConfigMap == nil {
 		path := filepath.Join(i.ManifestDir, "ixia-configmap.yaml")
 		if _, err := osStat(path); err != nil {
-			return fmt.Errorf("ixia configmap not found: %v", err)
+			log.Warnf("ixia controller deployed without configmap")
```
marcushines, line 828: why is this an option?

manifests/controllers/srlinux/crd/bases/kne.srlinux.dev_srlinuxes.yaml
```diff
@@ -0,0 +1,141 @@
+# Copyright 2022 Nokia
```
marcushines, line 1: there should be a way to generate all these yamls into a single file

alexmasi:
The presubmit will not test these changes because it uses our internal configs instead, not sure the best course of action here considering we would ideally like to test both internal and external configs

(and we would have to manually load our internal images in the test to test the external configs)

deploy/deploy.go
alexmasi: Because we can't provide a config map due to non open sourced images (well I guess we technically can but this step would always fail unless the user has the correct secret setup)

marcushines, line 828: maybe add the warning the example of how to deploy the configmap

marcushines (approved):

marcushines (approved):

PR #228, opened 2022-09-29
Nokia SR Linux with Openconfig services lab example

hellt:
Hi @alexmasi 

Here is a lab example featuring SR Linux nodes with enabled openconfig (gnmi, gnoi, gribi, p4rt) services.

As we discussed, I have created detailed documentation for this lab aiming at a public audience who likely sees KNE for the first time. The documentation is going to be hosted on learn.srlinux.dev website, you can see a live preview on my staging server for this lab here - https://hellt.github.io/learn-srlinux-stage1/tutorials/infrastructure/kne/srl-with-kne/

Once this PR is merged, I will move the docs from staging server to prod.

If you have any comments about the documentation provided for this lab, you may want to check this PR with source md content https://github.com/srl-labs/learn-srlinux/pull/38

PS. I removed stale examples from `srlinux` directory to keep things up to date


examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,62 @@
...
+name: "2-srl-ixr6"
+nodes: {
+    name: "srl1"
+    vendor: NOKIA
+    model: "ixr6e"
+    config:{
+        image: "ghcr.io/nokia/srlinux:22.6.3"
+        file: "22.6.3-opeconfig.cfg.json"
+    }
+    services:{
```
alexmasi, line 16: I read the instructions doc and it seems the other services (gnoi, gribi, p4rt) are also enabled by the config even though they are not specified here.

Ondatra reads the services section of the proto to determine which services are enabled and which ports they are available on. Additionally Ondatra determines the certificate type (self signed, etc.) for connecting to these services from the proto. Since these fields are not defined here we would need to do some post processing on the protobuf to add all of the exposed services and the cert type.

Just so I understand correctly some services (gRIBI, p4rt, etc.) are defined fully by the device config and cannot be specified here?

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,62 @@
...
+        key: 22
+        value: {
+            name: "ssh"
+            inside: 22
+        }
+    }
+    services:{
+        key: 57400
```
alexmasi, line 24: Ideally we could map these services and ports in the example to the openconfig "standard" ports:

```
  services: {
    key: 9337
    value: {
      name: "gnoi"
      outside: 9337
      inside: <YOUR CONTAINER gNOI PORT>
    }
  }
  services: {
    key: 9338
    value: {
      name: "gnsi"
      outside: 9338
      inside: <YOUR CONTAINER gNSI PORT>
    }
  }
  services: {
    key: 9339
    value: {
      name: "gnmi"
      outside: 9339
      inside: <YOUR CONTAINER gNMI PORT>
    }
  }
  services: {
    key: 9340
    value: {
      name: "gribi"
      outside: 9340
      inside: <YOUR CONTAINER gRIBI PORT>
    }
  }
  services: {
    key: 9559
    value: {
      name: "p4rt"
      outside: 9559
      inside: <YOUR CONTAINER P4RT PORT>
    }
  }
}
```

The outside and inside port specifications allow this remapping. The key value does not have to match the inside or outside value

alexmasi:
Also want to mention that I really appreciate the thorough docs write up. I'll take a look in more detail and maybe leave some feedback. Either way this will be really helpful for our users interested in srlinux

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
hellt: Oh neat. I will do that. I simply missed them :( 

One important question -- here you have gnoi defined over port 9337. I dont see this port to be assigned by IANA https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xml
Moreover, it says that gnoi-gnmi share the same 9339 port. Same is visible in the grpc-tunnel project that makes no distinction between the two protocols. 

Out implementation on srlinux also runs two servers on the same socket... So I wonder where does 9337 for gnoi comes from?

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
hellt: sorry, I missed that. Will add all enabled services.

hellt: added in [6a08c8c](https://github.com/openconfig/kne/pull/228/commits/6a08c8ccc9d208b446be5529f3c17b790b4bff35)

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
hellt: added missing services config using proposed port numbers in [6a08c8c](https://github.com/openconfig/kne/pull/228/commits/6a08c8ccc9d208b446be5529f3c17b790b4bff35)

note, that gnmi and gnoi runs over the same port as per my prev comment

alexmasi: Ah there was a typo is my example. The `gNxI` services can all have the same outside/inside ports and still have different service keys. When ondatra dials a service it checks for `gnoi` or `gnmi` etc.  in the service proto so it would be helpful to specify them each separately even if they point to the same ports

That should work out fine but let me know if there are any reported conflicts

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,90 @@
...
+# the startup config contains a self-signed TLS cert and has gnmi, gnoi, gribi, p4rt services enabled
+#
+# Read more about this topology at http://learn.srlinux.dev/tutorials/infrastructure/kne/srl-with-kne/
+
+name: "2-srl-ixr6"
+nodes: {
+    name: "srl1"
+    vendor: NOKIA
+    model: "ixr6e"
+    config:{
```
alexmasi, line 12: I see the cert field is left out in the proto, and its entirely defined by the config file. is the intention for the user to prefer config defined certs over proto defined certs? Is the node implementation of Certer intended to be used only in case where the user wants to create/delete new certs after startup without pushing config?

Just a general question so we can clearly communicate to users what is the best practice. Right now it is not entirely straightforward. I think the problem in general is the proto has to be looked at in combination with the config because proto fields can be empty while there are actually certs on the device

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
hellt: I was not able to use the same key for gnmi/gnoi. Either metallb or kne barked about key duplicates

alexmasi: the key needs to be unique but the ports can be the same, ondatra only looks at the service name

hellt: Thanks. I will have another look

hellt: @alexmasi I think I fixed that by carving out gnoi service in cd1148b

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
hellt: yes, the issue with self-signed certs that we have with srlinux today is that if you specify the self-signed certificated in the topo, then this cert is generated _after_ the node boots.

In order to use that certificate you have to create a binding between the TLS profile and services that intend to use it. When a certificate embedded in the config, you can provide all the bindings right away, without needing to configure them after the lab deployment.

When `srl-controller` adds `ConfigPush` you will be able to make use of the topo-defined certificates by pushing config that will make the binding between the self-generated cert and the services (gnmi, gnoi, etc) that want to use it.

Of course, you still can generate a self signed cert on srlinux and create the necessary bindings using gnmi, I just thought that packing it all in a config file is _good enough_, as far as self-signed certs are concerned.

alexmasi:
These changes look good to me. Since the presubmit will not check these services (not yet at least) I am going to run through the instructions and verify everything works in our environment. Then I will merge this

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,114 @@
+# This topology consists of two Nokia SR Linux IXR-6e nodes running 22.6.3 release and connected with a single link over ethernet-1/1 interfaces.
+# Each of the nodes comes up with a startup config which is sourced from 22.6.3-opeconfig.cfg.json file;
+# the startup config contains a self-signed TLS cert and has gnmi, gnoi, gribi, p4rt services enabled
+#
+# Read more about this topology at http://learn.srlinux.dev/tutorials/infrastructure/kne/srl-with-kne/
```
hellt, line 5: this link will be live when this PR is merged

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,114 @@
...
+
+name: "2-srl-ixr6"
+nodes: {
+    name: "srl1"
+    vendor: NOKIA
+    model: "ixr6e"
+    config:{
+        image: "ghcr.io/nokia/srlinux:22.6.3"
+        file: "22.6.3-opeconfig.cfg.json"
```
alexmasi, line 14: typo here

alexmasi:
This change also requires the a newer version of the srlinux controller then we have in the checked in manifests

Let me file a separate PR to roll that forward

examples/srlinux/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -0,0 +1,114 @@
...
+# Read more about this topology at http://learn.srlinux.dev/tutorials/infrastructure/kne/srl-with-kne/
+
+name: "2-srl-ixr6"
+nodes: {
+    name: "srl1"
+    vendor: NOKIA
+    model: "ixr6e"
+    config:{
+        image: "ghcr.io/nokia/srlinux:22.6.4"
+        file: "22.6.3-openconfig.cfg.json"
```
alexmasi, line 14: should be updated to match

hellt: I have renamed the file to `srl-openconfig.cfg.json` to make sure it can be used with future srl image versions without misguiding users as to which version it belongs to.

alexmasi (approved):

alexmasi (approved):

PR #231, opened 2022-10-04
added config push for Nokia SR Linux

hellt:
this PR adds `ConfigPush` method to Nokia SR Linux.

The bump to scrapligo was required since a fix was made to detect errors in multi response object.


topo/node/srl/srl.go
```diff
@@ -94,6 +99,38 @@ func (n *Node) GenerateSelfSigned(ctx context.Context) error {
 	return n.cliConn.Close()
 }
 
+// ConfigPush pushes config lines proivded in r using scrapligo SendConfig
```
alexmasi, line 102: typo

topo/node/srl/srl.go
```diff
@@ -94,6 +99,38 @@ func (n *Node) GenerateSelfSigned(ctx context.Context) error {
...
+func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
+	log.Infof("%s - pushing config", n.Name())
+
+	cfg, err := io.ReadAll(r)
+	cfgs := string(cfg)
+
+	log.Debugf("config to push:\n%s", cfgs)
```
alexmasi, line 109: not sure if we should log the entire cfgs content in case of a very large config. maybe its better to log the head of the config and the total size?

topo/node/srl/srl_test.go
```diff
@@ -310,3 +310,100 @@ func TestResetCfg(t *testing.T) {
...
 }
+
+func TestConfigPush(t *testing.T) {
+	ki := fake.NewSimpleClientset(&corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "pod1",
+		},
+	})
+
+	reaction := func(action ktest.Action) (handled bool, ret watch.Interface, err error) {
```
alexmasi, line 321: is the watch reactor required here?

topo/node/srl/srl_test.go
```diff
@@ -310,3 +310,100 @@ func TestResetCfg(t *testing.T) {
...
+				scrapliopts.WithTransportType(scraplitransport.FileTransport),
+				scrapliopts.WithFileTransportFile(tt.testFile),
+				scrapliopts.WithTimeoutOps(2 * time.Second),
+				scrapliopts.WithTransportReadSize(1),
+				scrapliopts.WithReadDelay(0),
+				scrapliopts.WithDefaultLogger(),
+			}
+
+			if scrapliDebug() {
+				li, _ := scraplilogging.NewInstance(
```
alexmasi, line 390: I'm assuming the ignored value is an error? If so it should not be ignored but instead logged using t.Log or warn or whatever to let the tester know that the scrapli logging could not be created (without needing to fail the test)

topo/node/srl/configpush_fail_cli.cfg
```diff
@@ -0,0 +1,2 @@
+set / system inftion location "wrong command"
```
alexmasi, line 1: this is such a small nit but I couldn't help myself from pointing out that the files are named configpush_{} and configpush_{}_cli.cfg and one is fail and the other is failure while success is consistent

hellt: good call, I missed that 
fixed in 63afc4c30438bfd7f72ced7952f4956b9d4e57ac

topo/node/srl/srl.go
hellt: fixed

topo/node/srl/srl_test.go
hellt: yes, you're right. I have removed it from the two tests that don't require it.

topo/node/srl/srl_test.go
hellt: I have added a check for errors with a test abort, as I think if the logging was required and it doesn't work, it makes sense to see what is wrong with the dependency

topo/node/srl/srl.go
hellt: I used the approach that currently exists in the repo.
In my opinion, observing the full config is beneficial as instructed by the user using the `-v debug` flag. The verbosity was requested.

Configs are indeed massive on SR Linux due to explicit ACLs, but when running without a debug they are not visible

alexmasi (approved):

PR #238, opened 2022-10-11
Add keysight manifest for deploy

marcushines:


alexmasi (commented):
the versions file should be named ixia-configmap.yaml: https://github.com/openconfig/kne/blob/1b4d2060101e3df472665f651c383efd322e0264/deploy/deploy.go#L826

but I like this pattern better, can we change it internally as well during the sync?

frasieroh:
can we temporarily accept both `ixia-configmap` and `ixia-version` and have them be equivalent and mutually exclusive? We have an internal repo that distributes our own manifests and images to keeps things stable, and I'd prefer not to have it only work with versions of KNE before or after this change.

manifests/keysight/ixiatg-version.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: v1
```
alexmasi, line 1: rename this file to ixiatg-configmap

deploy/kne/test.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
alexmasi, line 1: this file looks leftover from manual testing?

alexmasi:
the presubmit failed because this PR does not have the changes on the main branch that fix the issue

alexmasi (dismissed):

alexmasi (dismissed):

alexmasi (approved):

PR #244, opened 2022-10-17
Remove references to node type

alexmasi:


alexmasi:
The linter is complaining about use of the deprecated field:

```
Error: SA1019: impl.Proto.Type is deprecated: Do not use.  (staticcheck)
```

So its safe to ignore the failure

guoshiuan (approved):

PR #248, opened 2022-10-21
Ixia support for custom init container

anjan-keysight:
Ixia support for custom init container through KNE config files


topo/node/ixia/ixia.go
```diff
@@ -59,6 +59,11 @@ func (n *Node) newCRD() *ixiatg.IxiaTG {
 		},
 	}
 
+	if n.GetProto().Config != nil {
+		ixiaCRD.Spec.InitContainer.Image = n.GetProto().Config.InitImage
+		ixiaCRD.Spec.InitContainer.Sleep = n.GetProto().Config.Sleep
```
alexmasi, line 64: sleep should be the sleep time before starting the pod according to the documentation: https://github.com/openconfig/kne/blob/1775c426748ac4100544b08c9383709da53593fe/proto/topo.proto#L125

Does this instead create the pod without sleeping but then start the initcontainer after the sleep period?

alexmasi (dismissed):

alexmasi:
Looks like there are some merge conflicts, once those are fixed I can merge

anjan-keysight:
Resolved go.sum conflict.

alexmasi:
there are still merge conflicts unfortunately

anjan-keysight:
I have merged main and resolved the issue. Once this is merged I will sync the other PR and resolve conflicts.

alexmasi (approved):

PR #253, opened 2022-10-31
Add vendor data, update arista operator to v2.0.1

frasieroh:
- Add a vendor_data field to the node proto
- This allows the arista node to override feature toggles and block startup acknowledgement on certain internal processes.

- Define an arista-specific config type
- This is in proto/ceos and define APIs for the above features

- Upgrade to the v2.0.1 of the arista operator
- A new major version was needed for a breaking change related to testing. The existing uint32 fields did not serialize properly when using the dynamic fake, which was needed to test the CRD generation with vendor-specific data (and address a general coverage gap, too).

- Switch to the arista operator dynamic client
- Previously, we were using the REST client.


topo/node/ceos/ceos.go
```diff
@@ -152,10 +162,26 @@ func (n *Node) CreateCRD(ctx context.Context) error {
...
+			device.Spec.WaitForAgents = waitForAgents
+		}
+	}
 	// Post to k8s
-	client, err := ceosclient.NewForConfig(n.RestConfig)
-	if err != nil {
-		return err
+	var err error
+	// The client can be set by a test, for example, to a fake
+	if client == nil {
```
marcushines, line 180: i would recommend 

var (
 newClient = ceosclient.NewForConfig
)

then in your test you just replace with the fake.New...

then here you don't need the nil check you just call newClient()

frasieroh: hadn't thought of that, cool.

marcushines (approved):

PR #258, opened 2022-11-03
Remove the ios-xr/ prefix in GAR path.

guoshiuan:
Also rename e8000 to 8000e.


alexmasi (commented):
Can you change the cisco.go file as well to default to 8000e?

Also please change the directory structure of examples to 8000e as well

guoshiuan (commented):
Revised. PTAL.

alexmasi (approved):

PR #264, opened 2022-11-15
cptx: push gRPC config along with cert

nitinsoniism:
* make sure gRPC service starts after configuring certs
* modify UT
* remove gRPC config from default config
* fix hostname
* add cptx and host container image config

### DIFF
--- a/examples/juniper/cptx-host/cptx-host.pb.txt
+++ b/examples/juniper/cptx-host/cptx-host.pb.txt
@@ -5,7 +5,6 @@ nodes: {
     model: "cptx"
     os: "evo"
     config: {
-      image: "cptx:test"
       config_path: "/home/evo/configdisk"
       config_file: "juniper.conf"
       file: "juniper.config"
--- a/examples/juniper/cptx-host/cptx-host.pb.txt
+++ b/examples/juniper/cptx-host/cptx-host.pb.txt
@@ -5,12 +5,13 @@ nodes: {
     model: "cptx"
     os: "evo"
     config: {
+      image: "cptx:test"
       config_path: "/home/evo/configdisk"
       config_file: "juniper.conf"
       file: "juniper.config"
       cert: {
             self_signed: {
-                cert_name: "ca-ipsec",
+                cert_name: "grpc-server-cert",
                 key_name: "N/A",
                 key_size: 4096,
             }
@@ -71,6 +72,9 @@ nodes: {
 nodes: {
     name: "vm-1"
     vendor: HOST
+    config: {
+        image: "ubuntu:latest"
+    }
     services: {
         key: 22
         value: {
@@ -82,6 +86,9 @@ nodes: {
 nodes: {
     name: "vm-2"
     vendor: HOST
+    config: {
+        image: "ubuntu:latest"
+    }
     services: {
         key: 22
         value: {
--- a/examples/juniper/cptx-host/juniper.config
+++ b/examples/juniper/cptx-host/juniper.config
@@ -6,16 +6,6 @@ system {
         ssh {
             root-login allow;
         }
-        extension-service {
-            request-response {
-                grpc {
-                    ssl {
-                        hot-reloading;
-                        use-pki;
-                    }
-                }
-            }
-        }
     }
     syslog {
         file interactive-commands {
@@ -32,20 +22,6 @@ interfaces {
         unit 0 {
             family inet {
                 address FXP0ADDR;
-            }
-        }
-    }
-}
-openconfig-system:system {
-    openconfig-system-grpc:grpc-servers {
-        grpc-server grpc-server {
-            config {
-                services GNMI;
-                enable true;
-                port 32767;
-                transport-security true;
-                certificate-id ca-ipsec;
-                listen-addresses 0.0.0.0;
             }
         }
     }
--- a/topo/node/cptx/cptx.go
+++ b/topo/node/cptx/cptx.go
@@ -87,6 +87,21 @@ func (n *Node) SpawnCLIConn() error {
 	return err
 }
 
+// Returns config required to configure gRPC service
+func (n *Node) GetGrpcConfig() []string {
+	return []string{
+		"set system services extension-service request-response grpc ssl hot-reloading",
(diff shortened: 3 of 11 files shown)


examples/juniper/cptx-host/juniper.config
```diff
@@ -6,16 +6,6 @@ system {
         ssh {
             root-login allow;
         }
-        extension-service {
```
marcushines, line 9: why are you removing this config

examples/juniper/cptx-host/juniper.config
```diff
@@ -36,17 +26,3 @@ interfaces {
         }
     }
 }
-openconfig-system:system {
```
marcushines, line 39: same

topo/node/cptx/cptx.go
```diff
@@ -140,6 +140,26 @@ func (n *Node) GenerateSelfSigned(ctx context.Context) error {
...
+		"set system services extension-service request-response grpc ssl use-pki",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config services GNMI",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config enable true",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config port 32767",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config transport-security true",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config certificate-id grpc-server-cert",
+		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config listen-addresses 0.0.0.0",
+		"commit",
+	}
+	resp, err := n.cliConn.SendConfigs(cfgs)
```
marcushines, line 154: why are you sending this now as cli vs. in the initial config?

examples/juniper/cptx-host/cptx-host.pb.txt
```diff
@@ -5,12 +5,13 @@ nodes: {
     model: "cptx"
     os: "evo"
     config: {
+      image: "cptx:test"
```
marcushines, line 8: why add this?

examples/juniper/cptx-host/cptx-host.pb.txt
```diff
@@ -5,12 +5,13 @@ nodes: {
...
       config_path: "/home/evo/configdisk"
       config_file: "juniper.conf"
       file: "juniper.config"
       cert: {
             self_signed: {
-                cert_name: "ca-ipsec",
+                cert_name: "grpc-server-cert",
```
marcushines, line 14: why rename this?

examples/juniper/cptx-host/cptx-host.pb.txt
nitinsoniism: Just to keep this in sync with the other cptx example cptx-ixia/cptx-ixia.pb.txt

examples/juniper/cptx-host/cptx-host.pb.txt
nitinsoniism: ca-ipsec was not the apt name for this certificate as this is used for grpc

examples/juniper/cptx-host/juniper.config
nitinsoniism: Having this config part of the default config causes a race where gRPC service is started before the certer interface is invoked leading to failure. This should be configured after self-signed certs are configured.

topo/node/cptx/cptx.go
nitinsoniism: Having this config part of the default config causes a race where gRPC service is started before the certer interface is invoked leading to failure. This should be configured after self-signed certs are configured.

examples/juniper/cptx-host/cptx-host.pb.txt
alexmasi: Can you remove both this and the image field in cptx-ixia/cptx-ixia.pb.txt?

I would prefer to keep the examples using the node implementation defaults unless there is a specific version required to run the example (which in this case there is not)

nitinsoniism: makes sense. Done.

marcushines (dismissed):

nitinsoniism:
@marcushines @alexmasi could you please hold onto merging this for a bit? I need to clarify something before we merge this.

nitinsoniism:
I will send out another commit part of the same PR in a bit.

alexmasi:
Sounds good, I will also test this manually to make sure it works on our end as well before merging

nitinsoniism:
@alexmasi I am done with all my commits. You can try this out and let me know if you see any issues.

alexmasi:
Thanks, I'll give this a manual run tomorrow to confirm

alexmasi:
@nitinsoniism All looks good to me, however one question on the README. For gRIBI my output did not line up with the guide, I got:

```
$ gribic -a 192.168.11.50:9340 -u root -p Google123 --skip-verify get --ns default --aft ipv4
INFO[0000] target 192.168.11.50:9340: final get response:
INFO[0000] got 1 results
INFO[0000] "192.168.11.50:9340":
```

Is this a typo in the guide or is my config missing something?

nitinsoniism:
> @nitinsoniism All looks good to me, however one question on the README. For gRIBI my output did not line up with the guide, I got:
> 
> ```
> $ gribic -a 192.168.11.50:9340 -u root -p Google123 --skip-verify get --ns default --aft ipv4
> INFO[0000] target 192.168.11.50:9340: final get response:
> INFO[0000] got 1 results
> INFO[0000] "192.168.11.50:9340":
> ```
> 
> Is this a typo in the guide or is my config missing something?

Hey @alexmasi, thanks for trying this out. Yes, you will see a different result for gRIBI. This was just an example output I pasted in the README. The RIB entry was programmed using one of the ondatra scripts prior to running the gribic command. This wouldn't be part of the default config. One can use gribic too to program a RIB entry and it will display the same result.

alexmasi (approved):

PR #266, opened 2022-11-16
Add cptx + srl services coverage to presubmit

alexmasi:


alexmasi:
Need to update dep to latest ondatra commit once multicreds are supported in knebind

PR #267, opened 2022-11-17
Add 150 node topo

alexmasi:


examples/arista/ceos-150/arista.cfg
```diff
@@ -0,0 +1,67 @@
...
+no aaa root
+!
+username admin privilege 15 role network-admin secret sha512 $6$YDkKMqrZIRO65B7i$NfH7OT8ujV12S.5q2sH7c7l1AEqhYULt9OSf9/EKaC4S8cNLppS.otHhr0QlC9Kud9Xbt4OdxVOkZiwV/yBV/1
+!
+interface Ethernet1
+   no switchport
+!
+interface Ethernet1.100
+   encapsulation dot1q vlan 100
+   ip address 1.1.1.1/30
```
guoshiuan, line 31: Since we hardcode the IP address in the config, and all 150 routers use the same config, we don't expect the routers to work and this topology is only used to bring up them, correct?

alexmasi: Thats a good point. I wanted to create a very large topo for demo purposes (with many nodes and links). What I came up with is a link from node 1 -> 2, 2-> 3, ... , 149 -> 150, 150 -> 1

I would prefer not having to check in 150 router configs. I figured it would be okay to leave the hostname the same and enumerate the needed 2 interfaces. Traffic does not need to run in this setup. What should my config look like if this is my intended behavior? Currently I just copied the default arista one

guoshiuan (approved):

PR #274, opened 2022-12-05
Allow for '+' extensions to patch revision

Cerebus:
Should be pretty harmless, but lets me use vendored k8s releases on external clusters.


Cerebus:
I did the CLA dance, so damnfino why the rescan doesn't work.

alexmasi (commented):
can you add a unit test for the `getVersion` func?

Cerebus:
It would probably be better to replace the whole with https://pkg.go.dev/github.com/Masterminds/semver/v3 or another package like it.

PR #275, opened 2022-12-08
Lemming node impl uses operator

DanG100:
Note: not checking in a manifest because the images aren't released yet. No one is using lemming yet, so this shouldn't break anyone


alexmasi (commented):
looks good

can you update the `proto/controller.proto` and `controller/server/main.go` as well

DanG100:
done

alexmasi (approved):

PR #277, opened 2022-12-13
cptx: retry cert config when pki cert infra is not up

nitinsoniism:
At times even though the CLI is reachable doesn't mean the pki infra is up. If command handler or pki infra is not yet responding, retry a few times and bail out after a predefined delay.

### DIFF
--- a/topo/node/cptx/cptx.go
+++ b/topo/node/cptx/cptx.go
@@ -173,8 +173,7 @@ func (n *Node) GenerateSelfSigned(ctx context.Context) error {
 	}
 
 	// wait for cert infra to be ready and push certs
-	err = n.waitCertInfraReadyAndPushCert()
-	if err != nil {
+	if err := n.waitCertInfraReadyAndPushCert(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


topo/node/cptx/cptx.go
```diff
@@ -135,24 +172,10 @@ func (n *Node) GenerateSelfSigned(ctx context.Context) error {
...
-	commands := []string{
-		fmt.Sprintf("request security pki generate-key-pair certificate-id %s", selfSigned.GetCertName()),
-		fmt.Sprintf("request security pki local-certificate generate-self-signed certificate-id %s "+
-			"subject CN=abc domain-name google.com ip-address 1.2.3.4 email example@google.com",
-			selfSigned.GetCertName()),
-	}
-
-	multiresp, err := n.cliConn.SendCommands(commands)
+	// wait for cert infra to be ready and push certs
+	err = n.waitCertInfraReadyAndPushCert()
```
alexmasi, line 176: combine on to one line

alexmasi (dismissed):

topo/node/cptx/cptx.go
nitinsoniism: Done

alexmasi (approved):

PR #286, opened 2023-01-09
Accept all Ixia config map names

jasdeep-hundal:
Accept all Ixia config map names that are in use or to be used to ease the transition to just using ixiatg-version.yaml.

(Can add a test for individual file names, but I don't believe that actually exercises any different code paths from the existing test for a missing config map.)

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -901,13 +901,21 @@ func (i *IxiaTGSpec) Deploy(ctx context.Context) error {
 		return err
 	}
 	if i.ConfigMap == nil {
-		path := filepath.Join(i.ManifestDir, "ixiatg-configmap.yaml")
-		if _, err := osStat(path); err != nil {
+		var cmPath string
+		for _, cmName := range []string{"ixiatg-configmap.yaml", "ixia-configmap.yaml"} {
+			path := filepath.Join(i.ManifestDir, cmName)
+			if _, err := osStat(path); err == nil {
+				log.Infof("ixia controller configmap found at %q", path)
+				cmPath = path
+				break
+			}
+		}
+		if cmPath == "" {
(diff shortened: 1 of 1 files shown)


deploy/deploy.go
```diff
@@ -901,13 +901,22 @@ func (i *IxiaTGSpec) Deploy(ctx context.Context) error {
...
+			path := filepath.Join(i.ManifestDir, cmName)
+			if _, err := osStat(path); err != nil {
+				log.Warnf("ixia controller configmap not found at %q", cmName)
+				continue
+			}
+			cmPath = path
+			break
+		}
+		if cmPath == "" {
+			log.Warnf("ixia controller deployed without configmap, before creating a topology with ixia-c be sure to create a configmap following https://github.com/open-traffic-generator/ixia-c-operator#ixia-c-operator and apply it using 'kubectl apply -f ixiatg-version.yaml'")
```
alexmasi, line 915: ixiatg-confimap.yaml

deploy/deploy.go
```diff
@@ -901,13 +901,22 @@ func (i *IxiaTGSpec) Deploy(ctx context.Context) error {
 		return err
 	}
 	if i.ConfigMap == nil {
-		path := filepath.Join(i.ManifestDir, "ixia-configmap.yaml")
-		if _, err := osStat(path); err != nil {
-			log.Warnf("ixia controller deployed without configmap, before creating a topology with ixia-c be sure to create a configmap following https://github.com/open-traffic-generator/ixia-c-operator#ixia-c-operator and apply it using 'kubectl apply -f ixia-configmap.yaml'")
+		var cmPath string
+		for _, cmName := range []string{"ixiatg-version.yaml", "ixiatg-configmap.yaml", "ixia-configmap.yaml"} {
+			path := filepath.Join(i.ManifestDir, cmName)
+			if _, err := osStat(path); err != nil {
```
alexmasi, line 907: more useful to log which filepath was chosen, if none are found then the warning below is already enough

alexmasi (commented):
Marcus just submitted a PR, now `ixiatg-configmap.yaml` should be the preferred one. ixiatg-version can be removed. This change will definitely be helpful though for backward compatibility. Thanks

alexmasi (approved):

PR #288, opened 2023-01-11
Exclude presubmit test from makefile unit test cmd

alexmasi:


alexmasi:
Not triggering presubmit as its unaffected by makefile

bormanp (approved):

PR #289, opened 2023-01-13
Add logshim package in preparation for migrating from logrus to klog.

bormanp:
This package enables sending standard error output to a logger, one line at a time.

I expect to use it in deploy by setting the execer's stderr to a logshim that sends the lines to the logger of our choice.

### DIFF
--- a/logshim/logshim.go
+++ b/logshim/logshim.go
@@ -1,0 +1,113 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package logshim provides a shim to send an io.Writer to a logger, one line at
+// a time.  Normally only lines terminated with a newline are sent.  If a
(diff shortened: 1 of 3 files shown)


logshim/logshim.go
```diff
@@ -0,0 +1,115 @@
...
+	canceled chan struct{} // Closed when final data is flushed
+	write    func(...interface{})
+}
+
+const flushTimeout = time.Minute
+
+var testChannel chan struct{}
+
+// New returns an io.Writer that sends complete lines to the function f.  Each
+// complete line is written to f individually. Incomplete lines way up to one
```
alexmasi, line 41: may wait

logshim/logshim_test.go
```diff
@@ -0,0 +1,61 @@
...
+	if got := b.String(); got != want1 {
+		t.Errorf("First: Got %q, want %q", got, want1)
+	}
+
+	// Force the timer to go off
+	testChannel = make(chan struct{})
+	s.timer.Reset(1)
+	<-testChannel
+	if got := b.String(); got != want2 {
+		t.Errorf("First: Got %q, want %q", got, want1)
```
alexmasi, line 52: want2? is first correct here?

also want3 in below message?

alexmasi (approved):

logshim/logshim.go
bormanp: Done.  Thanks!

logshim/logshim_test.go
bormanp: Thanks.  Yeah, cut-n-paste error when I added the third case.   Would have been confusing if that error was produced!

alexmasi: Approved the change, just want to make sure that its intended to be First: prefix for two string comparison error messages. If that wasn't intended then I'll reapprove, otherwise feel free to merge this

alexmasi (approved):

PR #290, opened 2023-01-20
Migrate to klog from logrus

bormanp:
I will note that klog defaults to --logtostderr=true but in kne I have defaulted it to false as one reason to go to klog is to have real log files.

### DIFF
--- a/deploy/deploy.go
+++ b/deploy/deploy.go
@@ -52,7 +52,11 @@ var (
 	logOut        = os.Stderr
 	healthTimeout = time.Minute
 
-	execer execerInterface = kexec.NewExecer(logOut, logOut)
+	// Default standard out of children to log.Info
+	// and standar error to log.Warning.
+	execer execerInterface = kexec.NewExecer(
+		logshim.New(log.Info),
+		logshim.New(log.Warning))
 
 	// Stubs for testing.
 	execLookPath = exec.LookPath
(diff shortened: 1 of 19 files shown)


bormanp:
I should point out this is a user facing change, users will notice.  Are there documents we should modify?

deploy/deploy.go
```diff
@@ -510,7 +510,17 @@ func (k *KindSpec) setupGoogleArtifactRegistryAccess() error {
...
 	}
-	execer.SetStdout(log.StandardLogger().Out)
+	// It would be nice to something like:
+	//
+	//	errlog := logshim.New(func(args ...interface{}) {
+	//		log.InfoDepth(2, args...)
+	//	})
+	// So all the log lines don't just say they are coming from logshim, but
+	// it really doesn't help because the lines are coming from a goroutine
+	// that is using io.Copy.  So we just let the logs say logshim.
```
alexmasi, line 520: nit: maybe clean up this comment, not sure how much clarity it adds in its current state

deploy/deploy.go
```diff
@@ -48,10 +49,9 @@ data:
 )
 
 var (
-	logOut        = log.StandardLogger().Out
+	logOut        = os.Stderr
```
alexmasi, line 52: so the logs originally go to stderr until line 523?

shouldn't this line setup the logshim?

bormanp:
I have made the default to log to both files and stderr (otherwise it is silent).

alexmasi (approved):

PR #293, opened 2023-01-27
Print the list of logfiles created as the program terminates.

bormanp:
Switch --alsologtostderr to be false as --stderrthreshold=info will display all log lines upto and include INFO.

### DIFF
--- a/kne_cli/main.go
+++ b/kne_cli/main.go
@@ -17,7 +17,12 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
 
 	"github.com/openconfig/kne/cmd"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


kne_cli/main.go
```diff
@@ -40,9 +45,50 @@ func main() {
...
-		klog.Flush()
+	err := cmd.ExecuteContext(context.Background())
+	// err is referenced at the end of this function.
+	klog.Flush()
+
+	// klog provides no mechanism to determine where the logs are.  This
+	// code mimics the decsion that klog makes to determine the logging
+	// directory.
+
+	f := flag.Lookup("log_dir")
```
alexmasi, line 56: Thoughts on having another func `printLogFilenames` or similar to contain lines 57-89?

then the flow is a little more straightforward with the err val on line 48 being evaluated directly in line instead of needing the comment explaining that it is evaluated later

Not a big deal either way, code LGTM

alexmasi (dismissed):

alexmasi (approved):

PR #311, opened 2023-02-21
remove vendor implementations from coveralls

alexmasi:


alexmasi:
Overall coverage dropped, however now it reflects all pieces of code they we own and are responsible for coverage of

alexmasi:
Presubmit is a no op here so not going to trigger it

PR #317, opened 2023-03-02
Support specify manifests/operators as both raw data and filepaths

alexmasi:
This does not change the behavior of what is support with `kne deploy` (although a few yaml field names are changed).

This does however allow the KNE controller server to accept either raw bytes of a manifest/operator or a local filepath as previously supported. The IxiaTG controller also now accepts two parameters, configmap and operator so both can be changed independently.


guoshiuan (approved):

alexmasi:
This will fail the presubmit because the image that it will run against has old cfg in kne-internal, skipping

PR #318, opened 2023-03-04
Update gRIBI and gNMI ports to official numbers for lemming

wenovus:
per https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=gnmi

What else is missing to get the ports automatically to be these numbers when testing lemming? @DanG100

### DIFF
--- a/topo/node/lemming/lemming.go
+++ b/topo/node/lemming/lemming.go
@@ -194,15 +194,16 @@ func defaults(pb *tpb.Node) *tpb.Node {
 	}
 	if pb.Services == nil {
 		pb.Services = map[uint32]*tpb.Service{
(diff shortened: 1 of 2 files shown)


DanG100:
> per https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=gnmi
> 
> What else is missing to get the ports automatically to be these numbers when testing lemming? @DanG100

As of recently, Ondatra use kne as library, so you may be need to go mod replace kne with your local changes.

alexmasi (approved):

wenovus:
@alexmasi Could you merge it please? I don't have write permission.

PR #321, opened 2023-03-08
Add multivendor topology and valid configurations for ceos and xrd

marcushines:


examples/multivendor/multivendor.pb.txt
```diff
@@ -51,6 +97,42 @@ nodes: {
...
+  }
+  interfaces: {
+    key: "eth13"
+    value: {
+      name: "et-0/0/9"
+    }
+  }
 }
 nodes: {
   name: "srl"
```
alexmasi, line 138: xrd and nokia will need the extra service definitions including in their example textprotos

alexmasi (commented):
Just to confirm we want XRD here and not 8000e?

examples/multivendor/multivendor.pb.txt
marcushines: hrm?

alexmasi: nokia should look like this: https://github.com/openconfig/kne/blob/b091035fd078ad20f05ab4ea2444efad6ba36a51/examples/nokia/srlinux-services/2node-srl-ixr6-with-oc-services.pbtxt#L24

and xrd should be similar to 8000e I'd imagine: https://github.com/openconfig/kne/blob/b091035fd078ad20f05ab4ea2444efad6ba36a51/examples/cisco/8000e/8000e-ixia.pb.txt#L17

marcushines: done also found an issue where we can remove the outside setting requirement

topo/node/node.go
```diff
@@ -343,10 +343,13 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
 			name = fmt.Sprintf("port-%d", k)
 		}
+		if v.Outside != 0 {
+			log.Warningf("Outside should not be set by user. The key is used as the target external port")
+		}
 		sp := corev1.ServicePort{
 			Name:       name,
 			Protocol:   "TCP",
-			Port:       int32(v.Inside),
+			Port:       int32(k),
```
alexmasi, line 352: so to confirm the new behavior is:

port is key, if v.Outside is specified it overwrites it

targetPort is always v.Inside

should targetPort default to the key as well if v.Inside is unspecified? so the user could easily specify a service that uses the same inside and outside port:

```
services: {
  key: 22
  value: {
    name: "ssh"
  }
}
```

alexmasi (approved):

alexmasi (approved):

PR #322, opened 2023-03-08
New exec

bormanp:
New exec package that makes it easier for deploy to have a LogShim per execution so we will be able to add prefixes to to the log lines indicating what program the log information is coming from.


exec/exec_test.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		t.Run(tt.name, func(t *testing.T) {
+			c := Command(tt.cmd, tt.args...)
+			c.SetStdout(&stdout)
+			c.SetStderr(&stderr)
+			got := "nil"
+			if err := c.Run(); err != nil {
+				got = err.Error()
+			}
+			if !matches(got, tt.err) {
```
alexmasi, line 77: we use errdiff pkg (https://github.com/openconfig/gnmi/blob/master/errdiff/errdiff.go) quite a bit in this repo to simplify the err checking, I think it could simplify things here, what do you think?

exec/fake/fake.go
```diff
@@ -0,0 +1,296 @@
+// Package fake is used to fake calls github.com/openconfig/kne/exec.
+//
+// Typical Usage:
+//
+//	import "github.com/openconfig/kne/xec"
```
alexmasi, line 5: typo

bormanp: Done.

exec/exec_test.go
bormanp: I am actually one of the main authors of that!  :-)

Done.

exec/fake/fake_test.go
```diff
@@ -0,0 +1,417 @@
...
+func TestDone(t *testing.T) {
+	for _, tt := range []struct {
+		name string
+		send []Response
+		resp []Response
+		done string
+
+		// err is our expected "unexpected" error.
+		// This is used for all cases.
+		err error
```
alexmasi, line 332: this is not used or set in any test cases?

alexmasi (approved):

PR #326, opened 2023-03-10
srlinux: Use controller-runtime dynamic client instead of a generated clientset

n0shut:
Hi all,

It seems it is considered to be a common (good?) practice nowadays to use an API client provided by the controller-runtime package that allows users to work with both structured and unstructured objects while removing the need to generate clientsets [1] [2] [3] [4].

In this PR I am removing srl-controller clientset in favor of the controller-runtime client. Currently I put controller-runtime client under the Node struct, but if this approach will be considered good by the kne team it definitely makes sense to initialize a controller once for all node implementations. Each node implementation just needs to register its own schema with the client to make it all work.

This PR works with the tip of the branch that nokia works on in parallel - https://github.com/srl-labs/srl-controller/pull/31.

Looking forward to your review comments about this approach. 

/cc @marcushines 

[1] -- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.5/pkg/client
[2] -- https://hackernoon.com/platforms-on-k8s-with-golang-watch-any-crd-0v2o3z1q#h-interacting-with-cr-ds-the-new-way
[3] -- https://github.com/kubernetes-sigs/kubebuilder/issues/1152#issuecomment-584774562
[4] -- https://medium.com/@timebertt/kubernetes-controllers-at-scale-clients-caches-conflicts-patches-explained-aa0f7a8b4332#053d


n0shut:
I fake client exists from controller-runtime client -- https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake?utm_source=godoc

This allows to write unit tests for the CRs in question.

n0shut:
Hi @alexmasi @marcushines 
I have updated this PR to pass all KNE tests using the proposed change to srl-controller in https://github.com/srl-labs/srl-controller/pull/30.

I have also updated the srlinux example topology and startup file to feature srlinux 22.11.2 version and recent kind/kne versions.
For testing purposes I have released `srl-controller:0.5.0-rc1` controller container and matching manifests. I have installed it on a kind cluster with

```
kubectl apply -k https://github.com/srl-labs/srl-controller/config/default
``` 

Additionally, I verified that all OC services work based on the updated example topology. 

I would appreciate your comments and presubmit testing (granted you align the manifests to the new versions)

topo/node/srl/srl_test.go
```diff
@@ -62,6 +64,11 @@ func (f *fakeWatch) ResultChan() <-chan watch.Event {
 }
 
 func TestNew(t *testing.T) {
```
marcushines, line 66: func setUp() func() {
  origSrlinuxClient := newSrlinuxClient
  newSrlinuxClient = func(_ *rest.Config) (ctrlclient.Client, error) {
		return nil, nil
	}
return func() {
  newSrlLinuxClient = origSrLinuxClient
}

then your tests

defer setUp()()

n0shut:
Hi @alexmasi I updated the manifest, would be cool to see if we can run a presubmit using this branch

examples/nokia/srlinux-services/srl-openconfig.cfg.json
```diff
@@ -2361,9 +2401,6 @@
     }
   },
   "srl_nokia-network-instance:network-instance": [
-    {
-      "name": "DEFAULT"
-    },
```
alexmasi, line 2366: I would like this line to be reverted, helps align this topo with the featureprofiles defaults

topo/node/srl/srl.go
```diff
@@ -211,14 +238,7 @@ func (n *Node) Create(ctx context.Context) error {
...
 func (n *Node) Delete(ctx context.Context) error {
-	c, err := srlclient.NewForConfig(n.RestConfig)
-	if err != nil {
-		return err
-	}
-	err = c.Srlinux(n.Namespace).Delete(ctx, n.Name(), metav1.DeleteOptions{})
-	if err != nil {
-		return err
-	}
+	n.ControllerClient.Delete(ctx, &srlinuxv1.Srlinux{ObjectMeta: metav1.ObjectMeta{Name: n.Name()}})
```
alexmasi, line 241: check err?

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/scrapli/scrapligocfg v1.0.0
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
-	github.com/srl-labs/srl-controller v0.4.6
+	github.com/srl-labs/srl-controller v0.4.7-0.20230311230716-e1c96973d463
```
alexmasi, line 24: the container is v0.5.0 and the corresponding lib is v0.4.7, is that intended?

hellt: yes, just for testing inside this PR.
If tests are ok I will take both as 0.5.0

I spoke to Marcus and I will regenerate the manifests to use `:latest` container tag, so that manifest stays more-or-less static

topo/node/srl/srl.go
n0shut: done [3190719](https://github.com/openconfig/kne/pull/326/commits/31907198706d07dd671e7d1201e3b66348d8adf9)

examples/nokia/srlinux-services/srl-openconfig.cfg.json
n0shut: done [a5ecb13](https://github.com/openconfig/kne/pull/326/commits/a5ecb13dc037ad71e265feaa5e014e48499c83a2)

manifests/controllers/srlinux/manifest.yaml
```diff
@@ -374,27 +413,58 @@ spec:
...
+            drop:
+            - ALL
       - args:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
         - --leader-elect
         command:
         - /manager
-        image: ghcr.io/srl-labs/srl-controller:0.4.6
+        image: ghcr.io/srl-labs/srl-controller:latest
```
alexmasi, line 467: move this to 0.5.0

go.mod
alexmasi: yeah I tested on my machine, good to switch to 0.5.0

examples/nokia/srlinux-services/2node-srl-ixr6-with-oc-services.pbtxt
```diff
@@ -10,7 +10,7 @@ nodes: {
     vendor: NOKIA
     model: "ixr6e"
     config:{
-        image: "ghcr.io/nokia/srlinux:22.6.4"
+        image: "ghcr.io/nokia/srlinux:22.11.2"
```
alexmasi, line 13: just remove this line so it defaults to the :latest tag (also in the other nokia examples as well)

alexmasi:
Okay testing 0.4.7-xxxxxx commit with 0.5.0-rc worked on my machine. The presubmit in its current state will not pass, I am working on a better presubmit to test things like this

Once the versions are 0.5.0 and the comments are addressed I will get this merged

manifests/controllers/srlinux/manifest.yaml
n0shut: done

go.mod
n0shut: done

marcushines (approved):

alexmasi (approved):

PR #335, opened 2023-03-20
Misc lemming fixes

DanG100:
* new operator version
* expose more services by default
* test coverage


topo/node/lemming/lemming.go
```diff
@@ -99,14 +99,15 @@ func (n *Node) Create(ctx context.Context) error {
...
-				SelfSigned: lemmingv1.SelfSignedSpec{
+			dut.Spec.TLS = &lemmingv1.TLSSpec{
+				SelfSigned: &lemmingv1.SelfSignedSpec{
 					CommonName: tls.SelfSigned.CommonName,
 					KeySize:    int(tls.SelfSigned.KeySize),
 				},
 			}
 		}
 	}
+	log.Warningf("lemming spec: %+v", dut)
```
alexmasi, line 110: is this intended?

alexmasi (commented):
This will fail the presubmit because it will pull version 1 instead of 2 of the operator, this is known issue with the presubmit so I will ignore it. Did you test this locally to confirm?

topo/node/lemming/lemming.go
DanG100: nope, debugging leftover removed

DanG100:
> This will fail the presubmit because it will pull version 1 instead of 2 of the operator, this is known issue with the presubmit so I will ignore it. Did you test this locally to confirm?

yup, it's working for me and in our presubmit

alexmasi (approved):

PR #338, opened 2023-03-20
use exec.stream for internal import

alexmasi:


alexmasi:
ignore lint deprecation failure because exec.streamcontext is not available internally

bormanp (approved):

PR #339, opened 2023-03-21
The new load package reads and does some validation of deployment con…

bormanp:
…figs.

The load package has no direct knowledge of the Deployment or DeploymentConfig structures.

New controllers can be added without needing to modify the kne deployment code.

The load package can be used independently of the KNE CLI to validate deployment configurations.

Currently it does not validate that the images are present as this must be done from within the KNE cluster.

### DIFF
--- a/load/deploy_test.go
+++ b/load/deploy_test.go
@@ -1,0 +1,128 @@
+package load_test
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	cmddeploy "github.com/openconfig/kne/cmd/deploy"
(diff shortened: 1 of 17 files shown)


bormanp:
I am still verifying that KNE deploys correctly.  Creating the PR now to give a change to look at what I have done.

bormanp:
The golangci-lint failure is the known failure mentioned before.

load/deploy_test.go
```diff
@@ -0,0 +1,128 @@
+package load_test
```
alexmasi, line 1: will this package naming overlap when load topology is added?

alexmasi (approved):

load/deploy_test.go
bormanp: No, I plan to add the topology code into this same package.

PR #340, opened 2023-03-21
update presubmit

alexmasi:


cloudbuild/kne_test.sh
```diff
@@ -0,0 +1,115 @@
+#!/bin/bash
+# Copyright 2022 Google LLC
```
bormanp, line 2: 2023?

cloudbuild/kne_test.sh
```diff
@@ -0,0 +1,115 @@
...
+pushd "$HOME/kne/kne_cli"
+go build -o kne
+cli="$HOME/kne/kne_cli/kne"
+popd
+
+# Deploy a kind cluster
+pushd "$HOME"
+$cli deploy kne/deploy/kne/kind-bridge.yaml
+
+kubectl get pods -A
```
bormanp, line 36: So the only requirement is that kubectl does not exit with an error?

cloudbuild/kne_test.sh
```diff
@@ -0,0 +1,115 @@
...
+if ! docker exec kne-control-plane crictl images | grep "docker.io/networkop/init-wait"; then
+  echo "Unable to find preloaded image in cluster"
+  exit 1
+fi
+
+# Cleanup the kind cluster
+kind delete cluster --name kne
+
+# Create a kubeadm single node cluster
+sudo kubeadm init --cri-socket unix:///var/run/cri-dockerd.sock --pod-network-cidr 10.244.0.0/16
```
bormanp, line 87: Is this run by a person or automation?  If the later, how does the sudo work?

cloudbuild/vendors_test.sh
```diff
@@ -31,31 +31,30 @@ popd
 
 # Deploy a cluster + topo
 pushd "$HOME"
-$cli deploy kne-internal/deploy/kne/kind-bridge.yaml
-$cli create kne/cloudbuild/presubmit/topology.textproto
+$cli deploy kne/deploy/kne/kind-bridge.yaml
```
bormanp, line 34: Should there be at least a 'set -x' here?

alexmasi: there is on line 16

cloudbuild/kne_test.sh
alexmasi: I included these mainly for existing in the logs in case a user is debugging a presubmit failure, but in reality they aren't that helpful. I can remove them if you think they add noise. I don't want to validate the state because its not very flexible if dep cfg changes

cloudbuild/kne_test.sh
alexmasi: this was copy pasted from my manual instructions, deleting

cloudbuild/kne_test.sh
alexmasi: done

cloudbuild/kne_test.sh
bormanp: Okay, that is fine then.  You can leave them in.

bormanp (approved):

PR #341, opened 2023-03-22
Add new pods package for monitoring pod status

bormanp:


pods/pods.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+var testData []byte
+
+// GetPodStatus returns the status of the pods found in the supplied namespace.
+func GetPodStatus(namespace string) ([]PodStatus, error) {
+	var data []byte
+	if testData == nil {
+		var stdout bytes.Buffer
+		args := []string{"get", "pods", "-o", "json"}
```
alexmasi, line 46: should be able to avoid execing here using a kube client:

pods, err := kClient.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{})

pods/pods.go
```diff
@@ -0,0 +1,237 @@
...
+type PodStatus struct {
+	Name       string // name of pod
+	Namespace  string
+	Containers []ContainerStatus
+	Phase      string // current phase
+	Error      error  // If there was an error
+}
+
+type ContainerStatus struct {
+	Name    string // name of pod
```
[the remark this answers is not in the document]
bormanp: Done.

pods/pods.go
```diff
@@ -33,127 +89,125 @@ const (
...
-		d := json.NewDecoder(r)
-		for {
-			var pod corev1.Pod
-			if err := d.Decode(&pod); err != nil {
-				if err != io.EOF {
-					select {
-					case <-quit:
-					case ch <- PodStatus{Error: err}:
-					}
+		// seen keeps track of
```
alexmasi, line 153: finish comment, not sure what seen is tracking, would we not want the status to update even if we've seen the pod UID before, or maybe thats the event UID?

pods/testcmd/main.go
```diff
@@ -3,22 +3,26 @@
 package main
```
alexmasi, line 3: is this just a manual test script? it could be added as a new subcommand to kne_cli if it adds functionality that users might want

pods/pods.go
```diff
@@ -0,0 +1,213 @@
...
+		if !c.Equal(&q.InitContainers[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+// Values for Phase.  These constants are copied from k8s.io/api/core/v1 as a
+// convenience.
+const (
```
alexmasi, line 83: these arent used anywhere

pods/pods.go
alexmasi: bump

pods/pods.go
bormanp: No, they are available for the caller of the package so they don't have to go find the original ones.

pods/pods.go
bormanp: Overcome by events.

bormanp: Oh, the change wasn't pushed yet.

alexmasi (dismissed):

alexmasi (approved):

PR #367, opened 2023-05-02
config reset for cisco 8000e

mojiiba:
This pull adds reset config for cisco 8000e.


topo/node/cisco/cisco.go
```diff
@@ -434,6 +449,51 @@ func isNode8000eUp(ctx context.Context, req *rest.Request) bool {
...
+	n.cliConn, err = n.GetCLIConn(scrapliPlatformName, opts)
+	// TODO: add the following pattern in the scrapli/scrapligo/blob/main/assets/platforms/cisco_iosxr.yaml
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "ERROR")
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "% Failed")
+	return err
+}
+
+func (n *Node) ResetCfg(ctx context.Context) error {
+	if n.Proto.Model == ModelXRD {
+		return status.Errorf(codes.Unimplemented, "reset config is not implemented for cisco xrd node")
```
alexmasi, line 477: is there a timeline on xrd support?

topo/node/cisco/cisco.go
```diff
@@ -434,6 +449,51 @@ func isNode8000eUp(ctx context.Context, req *rest.Request) bool {
...
+func (n *Node) SpawnCLIConn() error {
+	opts := []scrapliutil.Option{
+		scrapliopts.WithAuthBypass(),
+		scrapliopts.WithTimeoutOps(scrapliOperationTimeout),
+	}
+	// add options defined in test package
+	opts = append(opts, n.testOpts...)
+	opts = n.PatchCLIConnOpen("kubectl", []string{"xr"}, opts)
+	if n.Proto.Model != ModelXRD {
+		opts = n.PatchCLIConnOpen("kubectl", []string{"telnet", "0", "60000"}, opts)
```
alexmasi, line 465: I will test this to make sure but looks to me like non-xrd could have multiple cli cmds appended to the opts

topo/node/cisco/cisco.go
```diff
@@ -59,6 +69,11 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
 
 type Node struct {
 	*node.Impl
+
+	cliConn *scraplinetwork.Driver
+
+	// scrapli options used in testing
+	testOpts []scrapliutil.Option
```
alexmasi, line 76: please add

```
// Add validations for interfaces the node provides
var (
	_ node.Resetter     = (*Node)(nil)
)
```

beneath this struct definition

topo/node/cisco/cisco.go
```diff
@@ -434,6 +449,51 @@ func isNode8000eUp(ctx context.Context, req *rest.Request) bool {
...
+	}
+	var err error
+	n.cliConn, err = n.GetCLIConn(scrapliPlatformName, opts)
+	// TODO: add the following pattern in the scrapli/scrapligo/blob/main/assets/platforms/cisco_iosxr.yaml
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "ERROR")
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "% Failed")
+	return err
+}
+
+func (n *Node) ResetCfg(ctx context.Context) error {
```
alexmasi, line 475: is there a follow up PR for pushconfig?

topo/node/cisco/cisco.go
mojiiba: we will need a change on container side, and we will try to address it after 8000e is finalized.

topo/node/cisco/cisco.go
mojiiba: yes, i am trying to push the code for config soon. Also, note that right now we push the specified config in topology file  when bringing up the router, and the reset config restore that config. so, even without push config you should be able to run the tests with simply doing reset config.

alexmasi (commented):
Okay I tested this PR manually and it looks good to me, will approve and merge after linter + comment is resolved

topo/node/cisco/cisco.go
```diff
@@ -434,6 +453,60 @@ func isNode8000eUp(ctx context.Context, req *rest.Request) bool {
...
+	opts = n.PatchCLIConnOpen("kubectl", []string{"xr"}, opts)
+	if n.Proto.Model != ModelXRD {
+		opts = n.PatchCLIConnOpen("kubectl", []string{"telnet", "0", "60000"}, opts)
+	}
+	var err error
+	n.cliConn, err = n.GetCLIConn(scrapliPlatformName, opts)
+	// TODO: add the following pattern in the scrapli/scrapligo/blob/main/assets/platforms/cisco_iosxr.yaml
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "ERROR")
+	n.cliConn.FailedWhenContains = append(n.cliConn.FailedWhenContains, "% Failed")
+	n.cliConn.OnClose = endTelnet
```
alexmasi, line 476: isn't telnet only used for 8000e? if so does this additional telnet end cmd cause any issues with xrd? there is no good way for me to test this because there are no XRD ops that utilize the CLI conn in this PR

mojiiba: you are right, the default onClose is enough for xrd and there is not need to end telnet. guarded the code.

alexmasi (approved):

PR #369, opened 2023-05-05
push config and generate self-signed for cisco 8000e

mojiiba:
This pull adds push config for cisco devices. It also remove bgp config from default config for 8000e since these type of config needs to be loaded by tests when is needed.   Further, we change the code to return unimplemented for GenerateSelfSigned for cisco nodes.

### DIFF
--- a/topo/node/cisco/cisco.go
+++ b/topo/node/cisco/cisco.go
@@ -564,6 +564,10 @@ func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
 	return resp.Failed
 }
 
+func (n *Node) GenerateSelfSigned(context.Context) error {
+	return status.Errorf(codes.Unimplemented, "certificate generation is not supported")
+}
+
 func init() {
 	node.Vendor(tpb.Vendor_CISCO, New)
 }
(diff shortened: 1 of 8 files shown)


topo/node/cisco/cisco.go
```diff
@@ -511,6 +512,62 @@ func (n *Node) ResetCfg(ctx context.Context) error {
...
+	}
+	if strings.ToLower(strings.Trim(lastLine, " ")) != "commit" {
+		processedCfg += "commit\n"
+	}
+	return processedCfg
+}
+
+func (n *Node) ConfigPush(ctx context.Context, r io.Reader) error {
+	if n.Proto.Model == ModelXRD {
+		return status.Errorf(codes.Unimplemented, "reset config is not implemented for cisco xrd node")
```
alexmasi, line 537: config push

topo/node/cisco/cisco.go
```diff
@@ -511,6 +512,62 @@ func (n *Node) ResetCfg(ctx context.Context) error {
...
+	}
+	if resp.Failed == nil {
+		log.Infof("%s - finished config push", n.Impl.Proto.Name)
+	}
+
+	return resp.Failed
+}
+
+func (n *Node) GenerateSelfSigned(context.Context) error {
+	return status.Errorf(codes.Unimplemented, "certificate generation is not supported")
```
alexmasi, line 568: can you add a short line about how it could be enabled (pushing the grpc stanza in the device cfg), if there is a link to cisco docs that would be great

alexmasi (commented):
Okay let me test this out manually

alexmasi (approved):

PR #371, opened 2023-05-05
Upgrade arista ceoslab operator to v2.0.2

alexmasi:
v2.0.2 supports multiple services (with different external ports) targeting the same internal port


alexmasi:
It appears v2.0.2 isn't working as expected:

```
$ kubectl get services -A
NAMESPACE                        NAME                                                         TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                                      AGE
...
ceos                             service-r1                                                   LoadBalancer   10.96.13.9      172.18.0.50   6030:30412/TCP,9340:31170/TCP,22:31985/TCP   6m9s
...
```

Only ext port 6030 from the below services is showing up:

gnmi:
  ext: 6030 int: 6030
gnoi:
  ext: 6031 int: 6030

Also neither are dial-able

alexmasi:
Moving forward with merging as there are no noticeable regressions, even if multi-services don't work

alexmasi:
filed PR to fix linter: https://github.com/openconfig/common-ci/pull/3

guoshiuan (approved):

PR #379, opened 2023-05-25
Add support for `openconfig/magna` to KNE.

robshakir:
Add support for `openconfig/magna` to KNE. `magna` is a lightweight OTG
implementation that can be used as a traffic generator. It is defined as a
separate node type than `HOST` as it needs to be separated in ONDATRA to
distinguish it as an ATE impementation.

```
commit 356945c967b2a5cba7e544753384225386c75f97
Author: Rob Shakir <robjs@google.com>
Date:   Thu May 25 11:30:57 2023 -0700

    Add testing for magna node definition.
    
     * (M) topo/magna/magna_test.go
       - Add testing for a new node definition.

commit c377555d04172eb196818ac67f611ad8a26c3fa2
Author: Rob Shakir <robjs@google.com>
Date:   Thu May 25 11:02:20 2023 -0700

    Add support for openconfig/magna containers to KNE.
    
     * (M) proto/topo.proto
       - Add MAGNA as a node vendor
     * (M) proto/*(_grpc)?.pb.go
       - Regenerate protobufs.
     * (M) topo/topo.go
       - Link in the magna node definition.
     * (M) topo/node/magna/magna.go
       - Create a node definition for magna.
```


proto/topo.proto
```diff
@@ -41,6 +41,7 @@ enum Vendor {
   GOBGP = 8;
   NOKIA = 9;
   OPENCONFIG = 10;
```
alexmasi, line 43: Unfortunately there is a bit of a strange pattern developing here where we have an openconfig vendor that is always assumed to be lemming

Generally the pattern so far has been Vendor should be the same across all models that a vendor provides, for example vendor CISCO, model XRD or model 8000e. There is a single node impl for each vendor that can handle the specific bring up based on the model field value

However in this case I believe ondatra uses vendor without model to determine if the device is an ATE or not so we probably cant follow the same pattern with the openconfig vendor

Probably makes most sense to consider openconfig not as a vendor but instead the vendor is Magna or Lemming. This current PR works towards that, optionally there can be a follow up from Daniel to deprecate the openconfig vendor and create a lemming vendor in the future. @DanG100 

What are your thoughts? @robshakir

topo/node/magna/magna.go
```diff
@@ -0,0 +1,72 @@
...
+func defaults(pb *tpb.Node) *tpb.Node {
+	if pb.Config == nil {
+		pb.Config = &tpb.Config{}
+	}
+	if len(pb.GetConfig().GetCommand()) == 0 {
+		pb.Config.Command = []string{
+			"/app/magna",
+			"-v=2",
+			"-alsologtostderr",
+			"-port=40051",
```
alexmasi, line 53: are these meant to be default services and self-signed certs? if so please add a TODO to expose those as such otherwise ondatra will not be able to resolve the gnmi port as its not encoded in the textproto unless expressed as a service

proto/topo.proto
robshakir: i dug a bit into ONDATRA, where it does look pretty baked that we assume that any model of vendor `KEYSIGHT` is an ATE. i.e., opposite to the "assume all of vendor `OPENCONFIG` is lemming" we say "any model of `KEYSIGHT` is an ATE".
 
we could change this, but I wonder whether there's value in doing this -- since today I'm not aware of any "vendor" other than `OPENCONFIG` that provides both ATEs and devices.

maybe we could, rather than adding `MAGNA` here, we could add `OPENCONFIG_ATE`, and then use model `MAGNA`. the `topo/node` implementation can then be the place where the model is used to know to spin up magna. a similar change can be made to `topo/lemming` to know of different `OPENCONFIG` DUT implementations.

topo/node/magna/magna.go
robshakir: ah, no problem -- yes, they're two services that are exposed. i'll just go and make these default services in this implementation.

robshakir: done.

proto/topo.proto
alexmasi: Discussed a bit with @bormanp today. Seems like there are 4 options here:
1) vendor is MAGNA and we switch OPENCONFIG vendor to LEMMING as a follow up
2) vendor is OPENCONFIG_ATE
3) vendor is OPENCONFIG and we change the kne solver to use a combo of model + vendor
4) vendor is OPENCONFIG and we add a ATE field to the topo proto to differentiate without needing to know the model

I would lean toward 3 if we can successfully modify the solver. If not I would be in favor of 2. Let's finalize this early next week after we have a bit more consideration. Sorry for the delay

alexmasi: in summary please move this magna implementation to the lemming.go file, differentiate based on the model value (if model == magna, model == lemming etc.) similar to cisco's approach for 8000e vs XRD. finally add the default value for the ondatra-role label

I will make an internal change to the solver to use the ondatra-role label and we can merge this change and the ondatra PR after that gets synced out

alexmasi:
This is blocked on https://github.com/openconfig/kne/pull/382

proto/topo.proto
robshakir: Apologies for the delay responding -- was OOO for a bit.

That sounds good -- I'll refactor this as you suggest.

robshakir:
@alexmasi PTAL! I think this works as you described above -- I made sure that `ondatra-role` is always set for both types, and this should support both. I haven't tested it actually creates a lemming or magna yet - and it does require that users that have vendor `OPENCONFIG` specify a model in their config to get a lemming. (@DanG100 / @wenovus are probably most concerned with this at this point :-))

topo/node/openconfig/openconfig.go
```diff
@@ -31,14 +33,41 @@ import (
...
+const (
+	// modelMagna is a string used in the topology to specify that a magna (github.com/openconfig/magna)
+	// ATE instance should be created.
+	modelMagna string = "MAGNA"
+	// modelLemming is a string used in the topology to specify that a lemming (github.com/openconfig/lemming)
+	// device instance should be created.
+	modelLemming string = "LEMMING"
+	// ondatraRoleLabel is a string used to specify whether an entity should be considered a DUT or an ATE
+	// by ONDATRA. OpenConfig is a vendor who supplies both.
+	ondatraRoleLabel string = "ondatra-role"
```
alexmasi, line 45: these 3 consts can be imported from the node package

alexmasi (commented):
LGTM, handing over to Daniel for review of lemming changes. Should we default to model lemming if none is provided to not break existing textprotos? @DanG100

alexmasi:
Also forgot to mention that all files besides openconfig{_test}.go can now be rolled back

robshakir:
> LGTM, handing over to Daniel for review of lemming changes. Should we default to model lemming if none is provided to not break existing textprotos? @DanG100

I'd encourage us to be more explicit here and require these to be updated. Given where lemming is in its deployment, I would expect that these references are probably internal to the OpenConfig org on GitHub :-)

robshakir:
> Also forgot to mention that all files besides openconfig{_test}.go can now be rolled back
> 
> The example (+presubmit) textprotos should also be updated, as of now the presubmit will fail

ACK - can you /gcbrun here so that I can see which presubmits fail? :-)

alexmasi:
> > LGTM, handing over to Daniel for review of lemming changes. Should we default to model lemming if none is provided to not break existing textprotos? @DanG100
> 
> I'd encourage us to be more explicit here and require these to be updated. Given where lemming is in its deployment, I would expect that these references are probably internal to the OpenConfig org on GitHub :-)

I don't believe many users outside of google are using lemming as of now so it should be a relatively small blast radius. We do have some internal textprotos checked in that would need updating. As adding a model before this change gets merged doesn't break anything I will handle those now.

alexmasi:
@wenovus Looping in Wen because Daniel is OOO this week

topo/node/openconfig/openconfig.go
robshakir: Done, thanks, I had implemented this before merging `main` so had missed these. Thanks for pointing them out.

robshakir:
OK, should be good-for-review.

alexmasi (dismissed):

cloudbuild/vendors/testbed.textproto
```diff
@@ -113,6 +113,7 @@ duts {
 duts {
   id: "lemming"
   vendor: OPENCONFIG
+  model: "LEMMING"
```
alexmasi, line 116: this is not required in the testbed file but is required in the topology textproto, that is what is causing the presub error

wenovus:
> @alexmasi PTAL! I think this works as you described above -- I made sure that `ondatra-role` is always set for both types, and this should support both. I haven't tested it actually creates a lemming or magna yet - and it does require that users that have vendor `OPENCONFIG` specify a model in their config to get a lemming. (@DanG100 / @wenovus are probably most concerned with this at this point :-))

Making this explicit sounds good to me, and I agree with Alex that it makes sense to make this breaking change earlier rather than later.

Changes to lemming-related code also LGTM.

cloudbuild/vendors/testbed.textproto
robshakir: ACK, apologies. Missed that. Fixed.

alexmasi (approved):

PR #387, opened 2023-06-16
Add event proto

alexmasi:
This will be used in anonymous usage reporting


proto/event.proto
```diff
@@ -0,0 +1,88 @@
...
+}
+
+message Topology {
+  repeated Node nodes = 1;
+  int64 link_count = 2;
+}
+
+message Node {
+  topo.Vendor vendor = 1;
+  string model = 2;
```
bormanp, line 85: I could see some objection to this when the model of the device is unreleased.   I think tags 2 through 4 probably should be optional.

proto/event.proto
```diff
@@ -0,0 +1,88 @@
...
+  string version = 3;
```
bormanp, line 86: How important are version and os?  Wouldn't these be determinable from the vendor/model?  There probably should be a comment justifying each piece of information gathered.

bormanp (changes requested):
I think there should be a comment that says identifying information must not be added to these structures (which is why link_count is included but the list of links is not).

proto/event.proto
alexmasi: Added comments. Version and OS provide much less value than knowing the vendor+model so I've removed them.

proto/event.proto
alexmasi: We can configure this through the KNE config file. By default all reporting will be set to false. Later we can added an enum in the cfg to fine tune what is shared

bormanp (approved):

PR #393, opened 2023-06-30
Add event publishing to kne_cli and controller server

alexmasi:


alexmasi:
Now all of the functionality is in place, cmd line flags can now be supplied with a config file using viper:

```
$ cat ~/.kne/config.yaml
report_usage: false
progress: true
```

This example sets the default values of the `--report_usage` and `--progress` flags. The hierarchy for flag values is as follows:

1. The value specified as a flag when calling the CLI
2. The value found in the kne config file
3. The default value specified using viper.SetDefault()
4. The default value specified using the flag default

alexmasi:
Next step is to break into 3 PRs:

1. Adding viper and restructure CLI
2. Adding usage pkg
3. Adding event publishing to deploy.go, topo.go, and the necessary flags to set them from the CLI and controller server

alexmasi:
need to merge from main after https://github.com/openconfig/kne/pull/396 is merged

deploy/deploy.go
```diff
@@ -146,7 +162,56 @@ type kubeVersion struct {
...
+	switch d.Ingress.(type) {
+	case *MetalLBSpec:
+		c.Ingress = epb.Cluster_INGRESS_TYPE_METALLB
+	}
+	switch d.CNI.(type) {
+	case *MeshnetSpec:
+		c.Cni = epb.Cluster_CNI_TYPE_MESHNET
+	}
+	for _, cntrl := range d.Controllers {
+		switch cntrl.(type) {
```
bormanp, line 183:
```
rantMode = true
defer func() { rantMode = false}()
```

While I understand why this is done and am not asking for it to be changed...

I really dislike this sort of switch.  Controllers should be able to report this themselves:

```
for _, cntrl := range d.Controllers {
    c.Controllers = append(c.Controllers, cntrl.Type())
}
```

Which would reduce the number of places that need changing when a new controller is added.  My guess is that this is not as easy because we are dealing with protobufs.

deploy/deploy.go
```diff
@@ -146,7 +162,56 @@ type kubeVersion struct {
...
+		case *LemmingSpec:
+			c.Controllers = append(c.Controllers, epb.Cluster_CONTROLLER_TYPE_LEMMING)
+		}
+	}
+	return c
+}
+
 func (d *Deployment) Deploy(ctx context.Context, kubecfg string) (rerr error) {
+	if d.ReportUsage {
+		r, err := metrics.NewReporter(ctx, d.ReportUsageProjectID, d.ReportUsageTopicID)
```
bormanp, line 199: Put the entire thing in a method so you can say:

```
// reportUsage returns a function to defer by the caller.
//
//      if d.ReportUsage {
//              finish = d.reportUsage(ctx)
//              defer finish(rerr)
//      }
func (d *Deployment) reportUsage(ctx context.Context) func(error) {
        r, err := metrics.NewReporter(ctx, d.ReportUsageProjectID, d.ReportUsageTopicID)
        if err != nil {
                log.Warningf("Unable to create metrics reporter: %v", err)
                return func(_ error) {}
        }
        if id, err := r.ReportDeployClusterStart(ctx, d.event()); err != nil {
                log.Warningf("Unable to report cluster deployment start event: %v", err)
                return func(_ error) { r.Close() }
        }
        return func(rerr error) {
                if err := r.ReportDeployClusterEnd(ctx, id, rerr); err != nil {
                        log.Warningf("Unable to report cluster deployment end event: %v", err)
                }
        }
}
```

topo/topo.go
```diff
@@ -146,9 +169,53 @@ func New(topo *tpb.Topology, opts ...Option) (*Manager, error) {
...
+	ReportCreateTopologyStart(context.Context, *epb.Topology) (string, error)
+	ReportCreateTopologyEnd(context.Context, string, error) error
+	Close() error
+}
+
 // Create creates the topology in the cluster.
-func (m *Manager) Create(ctx context.Context, timeout time.Duration) error {
+func (m *Manager) Create(ctx context.Context, timeout time.Duration) (rerr error) {
 	log.V(1).Infof("Topology:\n%v", prototext.Format(m.topo))
+	if m.reportUsage {
```
bormanp, line 202: See the above suggestion

deploy/deploy.go
alexmasi: yeah its not pretty

the other solution would be to extended the Controller interface to have a ControllerEventType() func with each spec implementing it with a method that returns the event enum. I wasn't really a fan of this solution either

deploy/deploy.go
alexmasi: done

topo/topo.go
alexmasi: done

README.md
```diff
@@ -84,6 +84,23 @@ number of production topologies.
 See the collection of [docs](docs/README.md) for in depth guides on how use
 Kubernetes Network Emulation (KNE).
 
+## Disclaimers
+
+### Usage Metrics Reporting
+
+We collect anonymous usage reporting metrics to gauge the health and
```
bormanp, line 91: As an initial sentence this may raise blood pressure.  Maybe something like:

The KNE CLI optionally collects anonymous usage reporting metrics. **This defaults to off.**
We use the metrics to gauge the health and ...

README.md
```diff
@@ -84,6 +84,23 @@ number of production topologies.
...
+performance of various KNE operations (i.e. cluster deployment, topology
+creation) on an **opt-in** basis. There is a global flag `--report_usage`
+that when provided shares anonymous details about certain KNE CLI commands.
+Collected data can be seen in the [event proto definition](proto/event.proto).
+**Usage metrics are not shared by default.** Additionally the PubSub project and
+topic the events are published to are configurable. If you want to track your
+own private metrics about your KNE usage then that is supported by providing a
+Cloud PubSub project/topic of your choosing. Full details about how/when usage
+events are published can be found in the codebase
+[here](kne/metrics/metrics.go). Usage metric reporting is appreciated to help
```
bormanp, line 101: We appreciate usage metric reporting as it helps us ...

deploy/deploy.go
bormanp: I feel better now :-)

bormanp (approved):

PR #395, opened 2023-07-06
Add config management with viper

alexmasi:


cmd/deploy/deploy.go
```diff
@@ -101,15 +97,12 @@ func deployFn(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return err
-	}
 	d, err := newDeployment(args[0], false)
 	if err != nil {
 		return err
 	}
-	if err := d.Deploy(cmd.Context(), kubecfg); err != nil {
+	d.Progress = viper.GetBool("progress")
+	if err := d.Deploy(cmd.Context(), viper.GetString("kubecfg")); err != nil {
```
bormanp, line 105: What happens if the GetString returns ""?  It would be nice if this (and possibly other) values could be sanity checked prior to running these functions as then you don't have to worry about checking the values.

kne_cli/main.go
```diff
@@ -45,7 +45,8 @@ func main() {
 		}
 	}
 	flags.Import()
-	err := cmd.ExecuteContext(context.Background())
+	root := cmd.New()
+	err := root.ExecuteContext(context.Background())
```
bormanp, line 49: You could just say `cmd.New().ExecuteContext(...)`

alexmasi: done

cmd/deploy/deploy.go
alexmasi: this is the same behavior as the current implementation, if the flag was the empty string then the function call would handle it as before. in the case of kubecfg the empty string is completely valid as an argument if the user wants to fallback to the in cluster config.

bormanp (approved):

PR #399, opened 2023-07-13
Event watcher for topo create

NehaManjunath:
Initial check-in

### DIFF
--- a/events/data_test.go
+++ b/events/data_test.go
@@ -1,0 +1,205 @@
+package events
+
+import (
+	"encoding/json"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Variables are named
+//
+//   name#data    Raw json from kubectl
+//   name#event     Data converted into a corev1.Event
+//   name#status  PodStatus version
+//   name#string  PodStatus as a string
+
+func json2event(j string) *corev1.Event {
+	var event corev1.Event
+	if err := json.Unmarshal(([]byte)(j), &event); err != nil {
+		panic(err)
+	}
+	return &event
+}
+
+var (
+	normal1event  = json2event(normal1eventdata)
+	warning1event = json2event(warning1eventdata)
+	warning2event = json2event(warning2eventdata)
+)
+
+// We only use 3 events when checking EventStatus.String as these cover all the cases.
+var (
+	normal1eventstring  = `{Name: "service-r2.1772fc5571f425ff", UID: "a756a650-bf61-4da6-8d69-d5ae40bd943c", Namespace: "ceos2"}`
+	warning1eventstring = `{Name: "r1.177351c307a38f38", UID: "a0450cc2-26e5-46e6-8ef8-be9dec02cd35", Namespace: "ceos1"}`
+)
+
+var (
+	normal1eventdata = `
+	{
+		"metadata": {
+			"name": "service-r2.1772fc5571f425ff",
+			"namespace": "ceos2",
+			"uid": "a756a650-bf61-4da6-8d69-d5ae40bd943c",
+			"resourceVersion": "3243",
+			"creationTimestamp": "2023-07-18T14:24:14Z",
+			"managedFields": [
+				{
+					"manager": "speaker",
+					"operation": "Update",
+					"apiVersion": "v1",
+					"time": "2023-07-18T14:24:14Z",
+					"fieldsType": "FieldsV1",
+					"fieldsV1": {
+						"f:count": {},
+						"f:firstTimestamp": {},
+						"f:involvedObject": {},
+						"f:lastTimestamp": {},
+						"f:message": {},
+						"f:reason": {},
+						"f:source": {
+							"f:component": {}
+						},
+						"f:type": {}
+					}
+				}
+			]
+		},
+		"involvedObject": {
+			"kind": "Service",
+			"namespace": "ceos2",
+			"name": "service-r2",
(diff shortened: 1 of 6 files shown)


events/status.go
```diff
@@ -0,0 +1,141 @@
...
+	cancel      func()
+	eventStates map[types.UID]string
+	ch          chan *EventStatus
+	stdout      io.Writer
+	warningf    func(string, ...any)
+
+	mu               sync.Mutex
+	progress         bool
+	currentNamespace string
+	currentEvent     types.UID
```
alexmasi, line 31: please fix

events/status.go
```diff
@@ -90,9 +90,9 @@ func (w *Watcher) Cleanup(err error) error {
 	select {
 	case werr := <-w.errCh:
 		if err != nil {
-			w.warningf("Deploy() failed: %v", err)
+			w.warningf("Create() failed: %v", err)
 		}
-		w.warningf("Deployment failed: %v", werr)
+		w.warningf("Topology creation failed failed: %v", werr)
```
alexmasi, line 95: failed appears twice

events/status.go
```diff
@@ -90,9 +90,9 @@ func (w *Watcher) Cleanup(err error) error {
 	select {
 	case werr := <-w.errCh:
 		if err != nil {
-			w.warningf("Deploy() failed: %v", err)
+			w.warningf("Create() failed: %v", err)
```
alexmasi, line 93: this is a generic watcher (as is the pods watcher), so this error message should be different

events/events.go
```diff
@@ -81,22 +81,17 @@ func WatchEventStatus(ctx context.Context, client kubernetes.Interface, namespac
 	}
 	kch := w.ResultChan()
 	ch := make(chan *EventStatus, 2)
+	initialTimestamp := metav1.Now()
 
 	go func() {
 		defer close(ch)
-		// seen is used to drop duplicate updates
```
alexmasi: do we not need de-deduping?

events/events.go
```diff
@@ -81,22 +81,17 @@ func WatchEventStatus(ctx context.Context, client kubernetes.Interface, namespac
 	}
 	kch := w.ResultChan()
 	ch := make(chan *EventStatus, 2)
+	initialTimestamp := metav1.Now()
```
alexmasi, line 84: should this also filter based on namespace in addition to time? There can multiple topos deployed in the same cluster only differentiated by namespace

alexmasi (commented):
Please fix all linter issues

events/events.go
```diff
@@ -0,0 +1,114 @@
+// Package events provides events status for a namespace using kubectl.
```
bormanp, line 1: It actually does not call kubectl (the original version of pods did, however)

events/events.go
bormanp: The namespace filtering is done by line 79

events/events.go
```diff
@@ -0,0 +1,114 @@
...
+func EventToStatus(event *corev1.Event) *EventStatus {
+	s := EventStatus{
+		Name:      event.ObjectMeta.Name,
+		Namespace: event.ObjectMeta.Namespace,
+		UID:       event.ObjectMeta.UID,
+	}
+	event.DeepCopyInto(&s.Event)
+	event = &s.Event
+	s.Type = event.DeepCopy().Type
+	s.Message = event.DeepCopy().Message
```
bormanp, line 112: You could just us event.Type and event.Message since you just copied this information into s.Event.

events/status.go
bormanp: In particular, something other than Create() might call it.

events/status.go
```diff
@@ -0,0 +1,137 @@
...
+// Watcher encountered an error the provided err is logged and the Watcher error
+// is returned, otherwise err is returned.
+func (w *Watcher) Cleanup(err error) error {
+	w.stop()
+	select {
+	case werr := <-w.errCh:
+		if err != nil {
+			w.warningf("Create() failed: %v", err)
+		}
+		w.warningf("Topology creation failed: %v", werr)
```
bormanp, line 94: It might not be a topology creation that is being watched.

events/events.go
alexmasi: aha, that was cutoff in my view. thanks

events/status.go
NehaManjunath: Fixed

events/status.go
NehaManjunath: thanks,fixed

events/status.go
NehaManjunath: Fixed to have a generic error message instead, thanks

events/events.go
NehaManjunath: Thanks, addressed this comment

events/status.go
NehaManjunath: Fixed to have a generic error message instead of create

events/events.go
NehaManjunath: We don't need deduping for events as there are no duplicates.

events/data_test.go
```diff
@@ -0,0 +1,218 @@
...
+import (
+	"encoding/json"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Variables are named
+//
+//   name#data    Raw json from kubectl
+//   name#event     Data converted into a corev1.Event
```
bormanp, line 12: Small amount of formatting needed

bormanp (approved):

PR #414, opened 2023-08-09
Use common workflow

wenovus:


wenovus:
@alexmasi Not sure why super-linter is returning "ERROR! Failed to call GitHub Status API!" So even though linting passed you don't see the GitHub statuses.

The concrete benefit here would be enabling build caching. Given that kne is somewhat different from the other repos I think it also makes sense for kne to maintain its own workflow, although I have added the necessary flags to make it work for KNE.

I'll just leave this PR here and let the maintainers decide whether using a common CI workflow is good for KNE.

alexmasi:
Closing in favor of https://github.com/openconfig/kne/pull/432 (which is the same but synced to head)

PR #415, opened 2023-08-09
Proto changes for host constraints

NehaManjunath:


proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
   // If interfaces is empty the interfaces defined in the links portion of the
   // topology will be populated into the node.
   map<string, Interface> interfaces = 12;
+  // Any host vm constraints required by the vendor.
```
alexmasi, line 84: I would expand this comment to indicate that these may not be required by a vendor, but it can be used to provide requirements for the node

"Any host constraints required by the node."

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
...
+  repeated HostConstraint host_constraints = 13;
+}
+
+// HostConstraint includes all the necessary system constraints required
```
alexmasi, line 88: HostConstraint is a constraint on the host where the topology is running.

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
...
   // topology will be populated into the node.
   map<string, Interface> interfaces = 12;
+  // Any host vm constraints required by the vendor.
+  repeated HostConstraint host_constraints = 13;
+}
+
+// HostConstraint includes all the necessary system constraints required 
+// which could be a sysctl setting/cgroup
+message HostConstraint{
+  oneof constraints{
```
alexmasi, line 91: I think this is better as singular

HostConstraint.GetConstraint() since each of these messages contains a single constraint only

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
...
+    // Includes the required sysctl settings that need to be configured
+    Sysctl sysctl_constraint = 1;
+  }
+}
+
+// Sysctl settings for the host
+message Sysctl{
```
alexmasi, line 98: I wonder if these are best described as KernelParameter or similar as sysctl is just the cli tool used to configure the kernel params

alexmasi (commented):
Looks good! left some comments mostly nits about the naming. After these are addressed lets have Paul take a look as well to see what he thinks about the design

proto/topo.proto
NehaManjunath: Addressed

proto/topo.proto
NehaManjunath: Thanks for the suggestion, fixed.

proto/topo.proto
NehaManjunath: Fixed.

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
...
+message HostConstraint{
+  oneof constraint{
+    // Includes the required sysctl settings that need to be configured
+    KernelParam sysctl_constraint = 1;
+  }
+}
+
+// Kernel parameter for Sysctl settings for the host
+message KernelParam{
+  // Actual name of the kernel parameter for sysctl setting
```
bormanp, line 99: // MIB style name of the kernel state to constrain.

From the manual page: The state is described using a ``Management Information Base'' (MIB) style name, listed in name, which is a namelen length array of integers.

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
...
+  string name = 1;
+  // Type of the sysctl setting which could be an array,
```
bormanp, line 101: This comment is wrong as currently the only constraint is a BoundedInteger.  I don't think you need a comment for the one of.  The comment for BoundedInteger below is sufficient.

proto/topo.proto
```diff
@@ -81,6 +81,34 @@ message Node {
   // If interfaces is empty the interfaces defined in the links portion of the
   // topology will be populated into the node.
   map<string, Interface> interfaces = 12;
+  // Any host constraints required by the node. These may not be required by a 
+  // vendor, but it can be used to provide requirements for the node.
+  repeated HostConstraint host_constraints = 13;
+}
+
+// HostConstraint is a constraint on the host where the topology is running.
```
bormanp, line 89: ... on the host where the node is running.

proto/topo.proto
NehaManjunath: Fixed.

proto/topo.proto
NehaManjunath: Fixed.

proto/topo.proto
NehaManjunath: Fixed.

proto/topo.proto
```diff
@@ -81,6 +81,32 @@ message Node {
...
+  // Any host constraints required by the node. These may not be required by a 
+  // vendor, but it can be used to provide requirements for the node.
+  repeated HostConstraint host_constraints = 13;
+}
+
+// HostConstraint is a constraint on the host where the node is running.
+message HostConstraint{
+  oneof constraint{
+    // Includes the required sysctl settings that need to be configured
+    KernelParam sysctl_constraint = 1;
```
alexmasi, line 93: I would consider changing the name sysctl_constraint to kernel_param_constraint or kernel_constraint

otherwise this LGTM

NehaManjunath: Addressed.

alexmasi (approved):

PR #420, opened 2023-08-21
juniper config mode retry and error handling

nitinsoniism:
In rare occasions, getting into config mode can take some time. Change here adds retry in case of privilege error by Scrapligo. Scrapligo throws this when config mode is not yet available. Also adds a bit better error handling and printing pod information when certificate push failure occurs.

Also gets rid of /sys/class mount which is no longer needed.

### DIFF
--- a/topo/node/juniper/juniper.go
+++ b/topo/node/juniper/juniper.go
@@ -115,7 +115,6 @@ func (n *Node) GRPCConfig() []string {
 
 // Waits and retries until CLI config mode is up and config is applied
 func (n *Node) waitConfigInfraReadyAndPushConfigs(configs []string) error {
-
 	log.Infof("Waiting for config to be pushed (timeout: %v) node %s", configModeTimeout, n.Name())
 	start := time.Now()
 	for time.Since(start) < configModeTimeout {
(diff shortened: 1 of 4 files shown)


topo/node/juniper/juniper.go
```diff
@@ -109,6 +113,38 @@ func (n *Node) GRPCConfig() []string {
 	}
 }
 
+// Waits and retries until CLI config mode is up and config is applied
+func (n *Node) waitConfigInfraReadyAndPushConfigs(configs []string) error {
```
alexmasi, line 117: Can you add a unit test that validates this new behavior?

Other than that this PR looks good

nitinsoniism: Done

topo/node/juniper/juniper_test.go
```diff
@@ -156,6 +156,20 @@ func TestGenerateSelfSigned(t *testing.T) {
 			ni:       ni,
 			testFile: "testdata/generate_certificate_failure",
 		},
+		{
```
alexmasi, line 159: please stub the sleep duration like line 133

nitinsoniism: Done. I see now test does not wait.

alexmasi (approved):

alexmasi:
Thanks!

PR #422, opened 2023-08-24
Validate host constraints

NehaManjunath:


topo/node/node.go
```diff
@@ -220,6 +225,66 @@ func (n *Impl) Create(ctx context.Context) error {
...
+}
+
+func (n *Impl) ValidateConstraints() []error {
+	hostConstraints := n.GetProto().GetHostConstraints()
+	if hostConstraints == nil {
+		return nil
+	}
+	var errorList []error
+	for _, hc := range hostConstraints {
+		switch v := hc.GetConstraint().(type) {
```
alexmasi, line 246: add a log.V(1) here for each cons that we are validating

topo/node/node.go
```diff
@@ -220,6 +225,66 @@ func (n *Impl) Create(ctx context.Context) error {
...
+	}
+	return combinedErrorString
+}
+
+func (n *Impl) ValidateConstraints() []error {
+	hostConstraints := n.GetProto().GetHostConstraints()
+	if hostConstraints == nil {
+		return nil
+	}
+	var errorList []error
```
alexmasi, line 244: use the errlist library instead, should make this easier: https://github.com/openconfig/kne/blob/fd11a07f35a7afb429ecc511400c1f53e5ad3d6b/deploy/deploy.go#L174

topo/node/cisco/cisco.go
```diff
@@ -405,6 +408,18 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		if pb.Config.Image == "" {
 			pb.Config.Image = "xrd:latest"
 		}
+		if pb.HostConstraints == nil {
+			pb.HostConstraints = append(pb.HostConstraints,
+				&tpb.HostConstraint{Constraint: &tpb.HostConstraint_KernelConstraint{
+					KernelConstraint: &tpb.KernelParam{Name: "fs.inotify.max_user_instances",
+						ConstraintType: &tpb.KernelParam_BoundedInteger{BoundedInteger: &tpb.BoundedInteger{
+							MaxValue: 64000, MinValue: 0}}}}})
```
alexmasi, line 416: I think this is a min cons, so leave max blank and it should default to infinity in the checking

topo/node/cisco/cisco.go
```diff
@@ -405,6 +408,18 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
...
+			pb.HostConstraints = append(pb.HostConstraints,
+				&tpb.HostConstraint{Constraint: &tpb.HostConstraint_KernelConstraint{
+					KernelConstraint: &tpb.KernelParam{Name: "kernel.pid_max",
+						ConstraintType: &tpb.KernelParam_BoundedInteger{BoundedInteger: &tpb.BoundedInteger{
+							MaxValue: 4194304, MinValue: 0}}}}})
```
alexmasi, line 421: can leave min unspecified

topo/node/cisco/cisco.go
```diff
@@ -405,6 +408,18 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		if pb.Config.Image == "" {
 			pb.Config.Image = "xrd:latest"
 		}
+		if pb.HostConstraints == nil {
```
alexmasi, line 411: also please confirm which constraint applies only to xrd or only to 8000e if that is the case

topo/node/cisco/cisco.go
```diff
@@ -413,6 +423,13 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		if pb.Config.Image == "" {
 			pb.Config.Image = "8000e:latest"
 		}
+		if pb.HostConstraints == nil {
```
alexmasi, line 426: hostconstraints will never be nil at this point

topo/node/node.go
```diff
@@ -220,6 +225,66 @@ func (n *Impl) Create(ctx context.Context) error {
...
+					errorList = append(errorList, fmt.Errorf("failed to validate kernel constraint error: %w", err))
+				}
+			}
+		}
+	}
+
+	return errorList
+}
+
+func validateBoundedInteger(protoConstraint *tpb.BoundedInteger, kernelConstraintInt int) error {
```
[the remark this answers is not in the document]
NehaManjunath: Fixed.

topo/node/node.go
NehaManjunath: Done.

topo/node/cisco/cisco.go
NehaManjunath: Fixed.

topo/node/cisco/cisco.go
NehaManjunath: Fixed.

topo/node/cisco/cisco.go
```diff
@@ -413,6 +438,13 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		if pb.Config.Image == "" {
 			pb.Config.Image = "8000e:latest"
 		}
+		if len(pb.HostConstraints) == 0 {
```
alexmasi, line 441: this condition will still never happen because of line 427

topo/node/node.go
```diff
@@ -220,6 +228,79 @@ func (n *Impl) Create(ctx context.Context) error {
...
+			}
+		default:
+			return nil
+		}
+	}
+
+	return errorList.Err()
+}
+
+var (
```
alexmasi, line 262: in the top var block I would define:

```
var (
    // Stubs for testing.
    kernelConstraintValue = kernelConstraintValueImpl
)

...

func kernelConstraintValueImpl(...) (int, error) {
    ...
}
```

This is a pattern we use for other stubs although nothing is wrong about your current approach.

topo/node/node.go
```diff
@@ -220,6 +228,79 @@ func (n *Impl) Create(ctx context.Context) error {
...
+	kernelConstraintValue = func(basePath string, constraint string) (int, error) {
```
alexmasi, line 263: why does this require a base path? Aren't all of the sysctl values found under the root dir?

topo/node/cisco/cisco.go
NehaManjunath: FIxed.

topo/node/cisco/cisco.go
NehaManjunath: line 427 is for model xrd. This line checks for hostconstraints in 8000e series based on a switch case. So this condition would happen, right?

topo/node/node.go
NehaManjunath: Thanks for the suggestion, modified accordingly.

topo/node/cisco/cisco.go
NehaManjunath: Resolved offline.

topo/node/node.go
```diff
@@ -100,6 +105,24 @@ var (
...
+
+func kernelConstraintValueImpl(constraint string) (int, error) {
+	constraintData, err := os.ReadFile(convertSysctlNameToProcSysPath(constraint))
+	if err != nil {
+		return 0, err
+	}
+	kcValue, err := strconv.Atoi(strings.TrimSpace(string(constraintData)))
+	if err != nil {
+		return 0, fmt.Errorf("failed to convert kernel constraint data: %s error: %w",
+			string(constraintData), err)
```
alexmasi, line 121: move these to the same line

topo/node/node.go
```diff
@@ -220,6 +246,64 @@ func (n *Impl) Create(ctx context.Context) error {
...
+	return errorList.Err()
+}
+
+// validateBoundedInteger - Evaluates a constraint if is within a bound of max - min integer. It defaults any unspecified upper bound to infinity,
+// the lower bound should already default to zero if not specified, validates that lower <= upper in the constraint otherwise error
+func validateBoundedInteger(nodeConstraint *tpb.BoundedInteger, hostCons int) error {
+	if nodeConstraint.MinValue > nodeConstraint.MaxValue {
+		return fmt.Errorf("invalid bounds. Max value %d is less than min value %d", nodeConstraint.MaxValue, nodeConstraint.MinValue)
+	}
+	if !nodeConstraint.ProtoReflect().Has(nodeConstraint.ProtoReflect().Descriptor().Fields().ByName(protoreflect.Name("max_value"))) {
```
alexmasi, line 286: i would just check if the field == 0 (the default value)

topo/node/cisco/cisco.go
```diff
@@ -405,6 +423,13 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		if pb.Config.Image == "" {
 			pb.Config.Image = "xrd:latest"
 		}
+		if len(pb.HostConstraints) == 0 {
```
alexmasi, line 426: i think checking for nil is better here in case it is unspecified

alexmasi (commented):
Okay, LGTM. Just 3 small nits and should be good to merge

topo/node/cisco/cisco.go
NehaManjunath: Addressed.

topo/node/node.go
NehaManjunath: My bad, had added it initially for testing. Removed it.

topo/node/cisco/cisco.go
NehaManjunath: Addressed. Constraints applicable per model:
xrd : fs.inotify.max_user_instances
8000e : kernel.pid_max

alexmasi (approved):

PR #432, opened 2023-09-19
add common github workflow

alexmasi:
credit to @wenovus


wenovus (approved):

.github/workflows/go.yml
```diff
@@ -8,28 +8,14 @@ on:
...
-          go-version: 1.20.x
-      - run: |
-          # shellcheck disable=2046
-          go test -v -coverprofile=profile.cov $(go list ./... | grep -v /cloudbuild) # ignore cloudbuild tests
-      - name: Send coverage
-        uses: shogo82148/actions-goveralls@v1
-        with:
-          path-to-profile: profile.cov
+  go:
+    uses: openconfig/common-ci/.github/workflows/go.yml@c2294c3c86c90e75e58c24a40d6f7f3364bbae9d
```
bormanp, line 12: Is this cherry picking a particular version of the go.yaml?

alexmasi: @wenovus

wenovus: Yes, this refers to the latest commit (minus the PR merge which created a new commit).

https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow recommends using a SHA for the best security and stability. Another reason to choose this over the alternative of tagging version v1.0.0, and pin on v1 to allow changes to common-ci to be auto-reflected is that it's very inconvenient to test changes to common workflows, so pinning to a particular commit pushes this responsibility to individual repo maintainers for when they need to upgrade their CI script.

bormanp (approved):

PR #449, opened 2023-10-25
Remove resetter interface from arista as this api doesn't

marcushines:
work in the default case with arista nodes currently


alexmasi:
lgtm, can you remove https://github.com/openconfig/kne/blob/main/cloudbuild/vendors_test.sh#L43 so that the presubmit tests the reset behavior?

alexmasi (approved):

PR #450, opened 2023-10-26
Support hardware loopback for ixiatg

alexmasi:


topo/node/keysight/keysight.go
```diff
@@ -322,6 +322,12 @@ func (n *Node) FixInterfaces() {
 	}
 }
 
+func (n *Node) HardwareLoopbackSupported() bool {
```
bormanp, line 325: After discussion with Alex we feel this is might not be the best name for this feature.  I am trying to think of better terminology but the only thing I have come up with us "BackToBack" vs "HardwareLoopback", but I am not at all set on that name.  Any other suggested names?  We can then look at the suggested names and select one.  LoopbackCable?  LoopbackLink?

bormanp (commented):
The code looks fine.

topo/node/keysight/keysight.go
alexmasi: - DirectLoopback
- SelfLoop
- SelfLoopback

I don't know what this situation is called on real hardware routers so I'll defer to you

alexmasi: settled on BackToBackLoop offline

bormanp (approved):

PR #454, opened 2023-10-30
Update Ixia manifests to 4555

jasdeep-hundal:


alexmasi (approved):

alexmasi:
The linter change is fixed in https://github.com/openconfig/kne/pull/450, it can be ignored

PR #463, opened 2023-11-27
Copying labels from KNE spec to POD spec.

alshabib:
This applies Juniper and Cisco vendors. Nokia
needs support for controller.


alexmasi:
can you add this to the default node impl as well? https://github.com/openconfig/kne/blob/4fe7a0fca59eaa50851596d527966fa234f484e4/topo/node/node.go#L395

PR #464, opened 2023-11-29
Add apply kubeyaml rpc to KNE grpc server

alexmasi:
Now kubeyaml can be applied to a cluster through the controller server


deploy/deploy.go
```diff
@@ -72,6 +72,27 @@ func logCommand(cmd string, args ...string) error {
 	return c.Run()
 }
 
+// logCommandWithInput runs the specified command but records standard output
+// with log.Info and standard error with log.Warning. in is sent to
+// the standard input of the command.
+func logCommandWithInput(in []byte, cmd string, args ...string) error {
```
bormanp, line 78: It would be nice to have a test for this code.  You should be able to test all three with a shell script as your input and sh as your command though you would probably need to have some test vars:

```
var logInfo = log.Info
var logWarning = log.Warning
```
so you can capture the output that would be sent to the log package.

deploy/deploy.go
```diff
@@ -429,6 +451,10 @@ func (e *ExternalSpec) GetDockerNetworkResourceName() string {
 	return e.Network
 }
 
+func (e *ExternalSpec) Apply(data []byte) error {
+	return logCommandWithInput(data, "kubectl", "apply", "-f", "-")
```
bormanp, line 455: This would be testable if you used a variable:

```
var kubectl = "kubectl"
```

and then in the test did something like `kubectl = "echo"`

cmd/deploy/deploy.go
```diff
@@ -43,6 +44,15 @@ func NewTeardown() *cobra.Command {
 	return teardownCmd
 }
 
+func NewApply() *cobra.Command {
```
bormanp, line 47: I don't mind adding the apply command if we also add testing for `applyFn`.  I would like to see our test coverage generally increase as we add new code.

deploy/deploy.go
```diff
@@ -72,6 +72,27 @@ func logCommand(cmd string, args ...string) error {
...
+	errLog := logshim.New(func(v ...interface{}) {
+		log.Warning(append([]interface{}{"(" + cmd + "): "}, v...)...)
+	})
+	defer func() {
+		outLog.Close()
+		errLog.Close()
+	}()
+	c.SetStdout(outLog)
+	c.SetStderr(errLog)
+	c.SetStdin(bytes.NewReader(in))
```
bormanp, line 92: Check if `len(in) > 0`

deploy/deploy.go
bormanp, line 78: rewrite logCommand in terms of logCommandWithInput:

```
func logCommand(cmd string, args ...string) error {
    return logCommandWithInput(nil, cmd, args...)
}
```

cmd/deploy/deploy.go
alexmasi: Decided to delete the command, we can decide at a later date how much of kubectl we want to wrap in KNE cli if any at all. Currently I would consider it a necessary tool to use along side the kne cli

deploy/deploy.go
alexmasi: done

deploy/deploy.go
alexmasi: good idea

alexmasi:
Adding the unit tests in a follow up commit to this PR

I intended to include them originally, wanted to make sure we were on the same page with the functionality and impl before writing them (I should have mentioned that in the description, sorry)

deploy/deploy.go
alexmasi: done

deploy/deploy.go
alexmasi: ack

deploy/deploy.go
```diff
@@ -51,57 +51,61 @@ var (
 	setPIDMaxScript = filepath.Join(homedir.HomeDir(), "kne-internal", "set_pid_max.sh")
 	pullRetryDelay  = time.Second
 	poolRetryDelay  = 5 * time.Second
+	logInfo         = log.Info
+	logWarning      = log.Warning
 )
 
+func runCommand(writeLogs bool, in []byte, cmd string, args ...string) ([]byte, error) {
```
bormanp, line 58: nice reduction in duplicated code!

bormanp (approved):

PR #479, opened 2024-01-22
Add credential refresh for GAR registries in kind

alexmasi:
Previously if GAR access was configured for kind clusters using `kne deploy`, the credential would expire after 1hr. This PR adds refresh before a new topology creation if extra GAR docker creds are found in the kind cluster.


alexmasi:
Still need to add the test cases for cluster/kind/kind_test.go

Other than that it should be ready for initial review

cluster/kind/kind.go
```diff
@@ -0,0 +1,172 @@
...
+	kubeletConfigPath  = "/var/lib/kubelet/config.json"
+)
+
+var (
+	kubeletConfigPathTemplate    = "%s:" + kubeletConfigPath
+	googleFindDefaultCredentials = google.FindDefaultCredentials
+	clusterKindNameRE            = regexp.MustCompile("kind-(.*)")
+)
+
+func ClusterIsKind() (bool, error) {
```
guoshiuan, line 28: See code dup in clusterKindName(). Can we call clusterKindName() to tell whether it is a kind cluster?

cluster/kind/kind.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to find gcloud credentials: %v", err)
+	}
+	token, err := creds.TokenSource.Token()
+	if err != nil {
+		return fmt.Errorf("failed to get token from gcloud credentials: %v", err)
+	}
+	for _, r := range registries {
+		s := fmt.Sprintf("https://%s", r)
+		if err := run.LogCommandWithInput([]byte(token.AccessToken), "docker", "login", "-u", "oauth2accesstoken", "--password-stdin", s); err != nil {
```
guoshiuan, line 106: Nit: Directly call 'fmt.Sprintf("https://%s", r)' as s here?

cluster/kind/kind.go
```diff
@@ -0,0 +1,172 @@
...
+	log.Infof("Setup GAR access for %v", registries)
+	return nil
+}
+
+func RefreshGARAccess(ctx context.Context) error {
+	nodes, err := clusterKindNodes()
+	if err != nil {
+		return err
+	}
+	cfg, err := run.OutCommand("docker", "exec", nodes[0], "cat", kubeletConfigPath)
```
guoshiuan, line 133: Is it because there is only one node? or we are always interested in the first node returned?

cluster/kind/kind_test.go
```diff
@@ -0,0 +1,373 @@
...
+			err := SetupGARAccess(ctx, tt.regs)
+			if s := errdiff.Substring(err, tt.wantErr); s != "" {
+				t.Fatalf("unexpected error: %s", s)
+			}
+		})
+	}
+}
+
+type fakeTokenSource struct {
+	tokenErr bool
```
guoshiuan, line 289: tokenErr reads to me an error instead a bool.. should we make it "hasTokenErr"?

cluster/kind/kind.go
alexmasi: I considered this, but clusterKindName treats a valid non-kind cluster as an error, whereas ClusterIsKind only errors on kubectl exec error. So technically this small dup is required

cluster/kind/kind.go
alexmasi: ack, prefer current for line length and readability

cluster/kind/kind_test.go
alexmasi: done

cluster/kind/kind.go
alexmasi: if one node doesnt have it, then none of them will, added a check and test case for no error but 0 nodes so thanks for pointing it out

guoshiuan (approved):

PR #484, opened 2024-01-24
Initial commit handling defaults in vendor node to make KNE topo create easier for users

NehaManjunath:


topo/node/arista/arista.go
```diff
@@ -366,10 +366,21 @@ func defaults(pb *tpb.Node) *tpb.Node {
...
+			pb.Model = "ceos"
+			pb.Labels["model"] = pb.Model
+		}
+	} else {
+		pb.Model = pb.Labels["model"]
 	}
+
 	if pb.Labels["os"] == "" {
-		pb.Labels["os"] = pb.Os
+		if pb.Labels["os"] = pb.Os; pb.Os == "" {
```
alexmasi, line 378: this can be simplified by defaulting pb.Os and pb.Model above this block. then labels can remain as is

topo/node/arista/arista.go
```diff
@@ -366,10 +366,21 @@ func defaults(pb *tpb.Node) *tpb.Node {
 		pb.Labels["vendor"] = tpb.Vendor_ARISTA.String()
```
alexmasi, line 366: for the services defaults above, I can see two desired behaviors:

1) a user wants to override to a custom gnmi port but still wants all of the other default services to come up so they specify only the gnmi change

2) a user wants a custom gnmi port but nothing else so they only specify gnmi

I think number 2 is less confusing so arista is correct, please make sure the other vendors follow that pattern as well (defaults are only added if services is nil)

topo/node/arista/arista.go
```diff
@@ -378,7 +389,18 @@ func defaults(pb *tpb.Node) *tpb.Node {
 		pb.Labels[node.OndatraRoleLabel] = node.OndatraRoleDUT
 	}
 	if pb.Config == nil {
-		pb.Config = &tpb.Config{}
+		pb.Config = &tpb.Config{
+			Image: "us-west1-docker.pkg.dev/gep-kne/arista/ceos:ga",
```
alexmasi, line 393: ceos:latest

topo/node/cisco/cisco.go
```diff
@@ -368,7 +368,18 @@ func defaults(pb *tpb.Node) (*tpb.Node, error) {
 		}
 	}
 	if pb.Config == nil {
-		pb.Config = &tpb.Config{}
+		pb.Config = &tpb.Config{
+			Image: "us-west1-docker.pkg.dev/gep-kne/cisco/xrd:ga",
+			Cert: &tpb.CertificateCfg{
```
alexmasi, line 373: cisco does not support this block, they are configured according to the cfg of the device

however adding it would be more accurate when looking at the proto even though its a no-op technically. can you look on a device and see what certname and keyname get generated?

topo/node/nokia/nokia.go
```diff
@@ -328,9 +328,6 @@ func (n *Node) Delete(ctx context.Context) error {
 }
 
 func defaults(pb *tpb.Node) *tpb.Node {
-	if pb.Config == nil {
```
alexmasi, line 331: some vendors add more or less labels, lets make it a standard to at least add labels for model, os, and vendor

topo/node/arista/arista.go
NehaManjunath: Fixed.

topo/node/arista/arista.go
NehaManjunath: Agreed, second option seems better, ensured that service defaults are added only if services is nil.

topo/node/nokia/nokia.go
NehaManjunath: Fixed. Added OS and Model info for all vendors.

topo/node/cisco/cisco.go
NehaManjunath: Ack. Found names "ems.pem" and "ems.key" under /xr-storage/config/grpc for certname and keyname. Defaulted to these values.

topo/node/arista/arista.go
```diff
@@ -359,6 +359,12 @@ func defaults(pb *tpb.Node) *tpb.Node {
 			},
 		}
 	}
+	if pb.Os == "" {
```
alexmasi, line 362: because arista uses a controller, are these values reflected in the final topo proto?

ideally we would leave defaults to them, but we also need the defaulted info to appear in the final textproto so we should prioritize that over controller defaults

PR #485, opened 2024-01-24
juniper cptx, ncptx allow inside gRPC port other than 32767

nitinsoniism:
This change helps allow users change inside port other than 32767. If user wants to configure a port other than default 32767, they need to modify the gnmi service in KNE protobuf config. In the absense of config the default 32767 will be used. Both reset to factory and cert generation will make sure configured port is used.

### DIFF
--- a/topo/node/juniper/juniper.go
+++ b/topo/node/juniper/juniper.go
@@ -39,6 +39,8 @@ var (
 	configModeTimeout = 10 * time.Minute
 	// Time between polls - config mode
 	configModeRetrySleep = 30 * time.Second
+	// Default gRPC port
+	defaultGrpcPort = uint32(32767)
 )
 
 const (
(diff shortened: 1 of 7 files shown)


topo/node/juniper/juniper.go
```diff
@@ -100,17 +100,21 @@ func (n *Node) SpawnCLIConn() error {
 
 // Returns config required to configure gRPC service
 func (n *Node) GRPCConfig() []string {
-	return []string{
+	port := n.Proto.Services[9339].GetInside()
```
alexmasi, line 103: it would be better to iterate through the service names in the values of the map to find "gnmi" or any of the g* services that run on the port rather than assuming 9339

topo/node/juniper/juniper.go
```diff
@@ -100,17 +100,21 @@ func (n *Node) SpawnCLIConn() error {
...
 		"set system services extension-service request-response grpc ssl use-pki",
 		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config services GNMI",
 		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config enable true",
-		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config port 32767",
 		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config transport-security true",
 		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config certificate-id grpc-server-cert",
 		"set openconfig-system:system openconfig-system-grpc:grpc-servers grpc-server grpc-server config listen-addresses 0.0.0.0",
-		"commit",
 	}
+	configs = append(configs, portConfig)
```
alexmasi, line 115: this func could remain mostly unmodified if `portConfig` is specified on line 111 and commit remains the same. if this new order is better then it should be `configs := []string{....}`

topo/node/juniper/juniper.go
nitinsoniism: 9339 is standard gNMI but included lookup in the services map just to be safe.

topo/node/juniper/juniper.go
nitinsoniism: Ahh, good catch. Didn't realize this can be done in Golang. Looks a lot cleaner. Got rid of other testing related change.

alexmasi (approved):

PR #489, opened 2024-02-01
Multiple services implementation

NehaManjunath:


topo/node/node.go
```diff
@@ -487,26 +487,31 @@ func (n *Impl) CreateService(ctx context.Context) error {
 		return nil
 	}
 	for k, v := range n.Proto.Services {
-		name := v.Name
-		if name == "" {
-			name = fmt.Sprintf("port-%d", k)
+		names := dedupServiceNames(append(v.Names, v.Name))
+		if names == nil {
```
bormanp, line 491: To be safe, this should be `if len(names) == 0` as a slice of zero length is not nil.

topo/node/node.go
```diff
@@ -535,6 +540,19 @@ func (n *Impl) CreateService(ctx context.Context) error {
 	return nil
 }
 
+func dedupServiceNames(s []string) []string {
+	allNames := make(map[string]bool)
+	deDuped := []string{}
+	for _, name := range s {
+		if !allNames[name] {
```
bormanp, line 547: if name == "" { continue }

topo/node/node.go
```diff
@@ -487,26 +487,31 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
-			Port:       int32(k),
-			TargetPort: intstr.FromInt(int(v.Inside)),
-		}
-		if v.NodePort != 0 {
-			sp.NodePort = int32(v.NodePort)
-		}
-		if v.Outside != 0 {
-			sp.Port = int32(v.Outside)
+		for _, name := range names {
+			if name == "" {
```
bormanp, line 495: This should be done in dedupServiceNames

topo/node/node.go
```diff
@@ -487,26 +487,31 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+				continue
+			}
+			if v.Outside != 0 {
```
bormanp, line 498: I think this can be done outside of the loop since you only have 1 v.

bormanp (changes requested):

topo/node/node.go
NehaManjunath: Ack.

topo/node/node.go
NehaManjunath: Fixed.

topo/node/node.go
NehaManjunath: Done.

topo/node/node.go
```diff
@@ -535,6 +537,19 @@ func (n *Impl) CreateService(ctx context.Context) error {
 	return nil
 }
 
+func dedupServiceNames(s []string) []string {
+	allNames := make(map[string]bool)
+	deDuped := []string{}
+	for _, name := range s {
+		if name != "" && !allNames[name] {
```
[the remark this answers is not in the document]
NehaManjunath: Thanks for the suggestion, incorporated the comment. Although I did have to change the deDuped variable from []string to string with separator "-" between service names.

topo/node/node.go
```diff
@@ -487,26 +487,28 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
-		if v.Outside != 0 {
-			sp.Port = int32(v.Outside)
+		for _, name := range names {
+			sp := corev1.ServicePort{
+				Name:       name,
+				Protocol:   "TCP",
+				Port:       int32(k),
+				TargetPort: intstr.FromInt(int(v.Inside)),
+			}
+			if v.NodePort != 0 {
```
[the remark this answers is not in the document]
NehaManjunath: Thanks for the detailed options and suggestions. Resolved and fixed issues that arose during testing.

topo/node/arista/arista.go
```diff
@@ -161,11 +161,11 @@ func (n *Node) CreateCRD(ctx context.Context) error {
 	for label, v := range proto.GetLabels() {
 		device.ObjectMeta.Labels[label] = v
 	}
-	for _, service := range proto.GetServices() {
+	for k, service := range proto.GetServices() {
 		if device.Spec.Services == nil {
 			device.Spec.Services = map[string]ceos.ServiceConfig{}
 		}
-		device.Spec.Services[service.Name] = ceos.ServiceConfig{
+		device.Spec.Services[node.DedupServiceNames(service, k)] = ceos.ServiceConfig{
```
bormanp, line 168: Can we address the service name before it is sent to the vendor code?  When the topology is imported we can check to see if service.Name is set, if not, check to see if there are any names and if so use the first one, and finally, if not, generate a name.  Then code like this does not have to change.  Or are you already working on this and I am just out of sequence?

bormanp (changes requested):

topo/node/arista/arista.go
NehaManjunath: Yes I was working on it, fixed it to handle when topology is imported so that vendor code need not be changed.

topo/topo_test.go
```diff
@@ -1944,3 +1944,69 @@ func TestStateMap(t *testing.T) {
...
+		{
+			desc: "services valid names with empty string",
+			svc: &tpb.Service{
+				Name:   "gnmi",
+				Names:  []string{"", "gribi"},
+				Inside: 9339,
+			},
+			want: &tpb.Service{
+				Name:   "gnmi",
+				Names:  []string{"", "gribi"},
```
bormanp, line 1989: Do we want to keep empty strings?  Is there any reason we should not strip them out?

NehaManjunath: Ack, stripped empty names.

alexmasi (approved):

topo/topo.go
```diff
@@ -594,16 +594,20 @@ func (m *Manager) push(ctx context.Context) error {
...
-	}
+	i := 0
 	for _, name := range s.Names {
 		if name != "" {
-			s.Name = name
-			return
+			s.Names[i] = name
+			i++
 		}
 	}
```
bormanp, line 603: I don't think this will work quite right.  The last element(s) can be repeated.  It is best to just create a new slice and then assign s.Names to that slice.

bormanp (changes requested):

topo/topo.go
NehaManjunath: Clarified the working of the snippet.

bormanp (approved):

PR #491, opened 2024-02-02
Add config support for alpine

DanG100:


topo/node/alpine/alpine.go
```diff
@@ -38,6 +39,19 @@ func New(nodeImpl *node.Impl) (node.Node, error) {
 	n := &Node{
 		Impl: nodeImpl,
 	}
+	idx := 1
```
alexmasi, line 42: this might make sense to move to its own func

DanG100: done

alexmasi (approved):

PR #497, opened 2024-02-16
Work towards cptx deprecation

alexmasi:
This PR changes the default of juniper from cptx to ncptx.

Also remove cptx from our presubmit

Additionally change the multivendor example to use ncptx so a nested virt enabled host is not required.


NehaManjunath (approved):
LGTM

PR #506, opened 2024-03-20
Update manifest.yaml to new version

DanG100:


alexmasi (approved):

alexmasi:
Linter is broken rn so you can ignore that

PR #523, opened 2024-03-26
Update cdnos controller

davnerson-dn:


alexmasi:
Thanks for the PR.

Can you also update the go.mod https://github.com/openconfig/kne/blob/5e92c292fd3845b1a13999a47575cf40e98fa44d/go.mod#L61 and move it to the upper require block?

alexmasi (approved):

PR #535, opened 2024-05-06
Add documentation on the expectations of vendor images.

bormanp:
Add documentation on the expectations of vendor images.

### DIFF
--- a/docs/vendor.md
+++ b/docs/vendor.md
@@ -1,0 +1,85 @@
+# Vendor Image Requirements
+
+A Vendor Image is a docker container that can be used with KNE to emulate a
+vendor's devices.
+
+Without vendor images KNE is just an empty virtual machine rack that does
+nothing.  Vendor supplied images are what the user of KNE sees and is interested
+in.  This document describes the requirements and expectations of vendor images
+to be used with KNE.
+
+This document describes the image requirements and expectations in order to be
+considered *KNE Qualified*.
+
+## KNE Uses
+
+KNE was built to enable testing the functionality of networks without physical
+hardware.  Due to the obvious limitations of emulation, KNE is not designed to
+test bandwidth and latency of connections.  KNE is designed to enable testing of
+the control protocols and interaction between devices.  There are several
+different types of testing.
+
+### Testing new Topologies
+
+KNE is used to test changes in network topology.  Changes in network topology
+can impact various protocols use in the network (e.g. BGP).
+
+### Testing Changes in Protocol or Configuration
+
+KNE is used to test protocol changes or other configuration changes.
+
+### Testing Device Functionality
+
+KNE is used to test changes to a device's Network Operating System (NOS).  This
+is a crucial step in validating a devices usability for a particular purpose
+when a new NOS is released.
+
+## Fidelity
+
+A network device in KNE can be viewed as two main components, the control plane
+and the data plane (the ASIC).
+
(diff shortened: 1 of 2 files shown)


docs/vendor.md
```diff
@@ -0,0 +1,85 @@
...
+found at <INSERT LOCATION HERE>.  It is expected that the image undergoes
+repeated testing to identify non-deterministic errors.
+
+## Support
+
+Vendors are responsible for support of their images.  This includes the node
+implementation in
+[kne/topo/node](https://github.com/openconfig/kne/tree/main/topo/node) as well
+as the vendor specific examples in
+[kne/examples](https://github.com/openconfig/kne/tree/main/examples).
```
chrisy, line 85: Are community contributions to vendor `node` implementations welcome? Do you expect or require vendors to provide review on community contributions?

bormanp: I think community contributions should be welcome though the vendor should be the one to approve changes to code they are responsible for.  I will updated the document.

Thanks!

docs/vendor.md
```diff
@@ -0,0 +1,87 @@
...
+A Vendor Image is a docker container that can be used with KNE to emulate a
+vendor's devices.
+
+Without vendor images KNE is just an empty virtual machine rack that does
+nothing.  Vendor supplied images are what the user of KNE sees and is interested
+in.  This document describes the requirements and expectations of vendor images
+to be used with KNE.
+
+This document describes the image requirements and expectations in order to be
+considered *KNE Qualified*.
```
alexmasi, line 12: what does KNE qualified mean in this case? Will there be an indicator anywhere that the image is KNE qualified, as in, is this the requirements for us to merge a new node impl? Or do mark some of the node impls as "qualified"?

docs/vendor.md
```diff
@@ -0,0 +1,87 @@
...
+* OpenConfig paths that report different results compared to the hardware.
+* Known limitations of the emulated device
+
+The listed OpenConfig paths need not be leaf nodes.  Wildcards may be used in
+the path where applicable.
+
+## Testing
+
+Vendor images must be tested prior to publication.  A standard set of tests is
+found at <INSERT LOCATION HERE>.  It is expected that the image undergoes
```
alexmasi, line 76: which tests are intended here?

docs/vendor.md
chrisy: I agree in principle, but I would caution about being too dogmatic about this. For example, I have a vendor who is not inclined to provide support for one containerized platform of theirs in KNE; I have local (not ready to publish, but do intend to) modifications to provide that support and I would be concerned if I seemed there was limited hope of this being accepted upstream given the vendors' apparent reticence.

bormanp: I think the question will come down to who has taken the responsibility to maintain that code.  If the ACME Network Switch company has provided a node implementation (implying they are the maintainer of that code) for most of their switches and you want to add in code to support one of the missing switches you have two paths.  The best option is to provide a PR that updates their node implementation and get their approval (as they maintain that node implementation).  If that path is not possible there can always be a new node implementation offered, though the benefits of this would need to out weigh the costs and technical debt.  If a the maintainer of a node implementation is non-response then you could take over maintenance of that code.

bormanp:
@chrisy @alexmasi PTAL

chrisy (approved):

PR #545, opened 2024-07-02
Add the ability to provide pre-generated certs

alexmasi:
Currently self-signed is the only cert type provided, this PR adds another option to the topo proto but does not yet add support to node impls to use the new field.

A follow up PR will use the field.


chrisy:
This has been top of mind for me, so thanks for tackling it. Generating certs is the slowest part of topology creation in many cases!

PR #550, opened 2024-07-18
Set a maximum amount of time to wait for an image to be pulled

bormanp:
after seeing ErrImagePull.

### DIFF
--- a/pods/status.go
+++ b/pods/status.go
@@ -14,6 +14,10 @@ import (
 	log "k8s.io/klog/v2"
 )
 
+// MaxPullTime is how long we will give a pod to start after seeing
+// ErrImagePull.
+var MaxPullTime = time.Minute * 15
+
 // A Watcher watches pod and container updates.
 type Watcher struct {
 	ctx       context.Context
@@ -21,6 +25,7 @@ type Watcher struct {
 	wstop     func()
 	cancel    func()
 	podStates map[types.UID]string
+	podStart  map[types.UID]time.Time
 	cStates   map[string]string
 	ch        chan *PodStatus
 	stdout    io.Writer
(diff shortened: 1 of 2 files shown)


pods/status.go
```diff
@@ -14,13 +14,18 @@ import (
 	log "k8s.io/klog/v2"
 )
 
+// MaxPullTime is how long we will give a pod to start after seeing
+// ErrImagePull.
+var MaxPullTime = time.Minute * 15
```
chrisy, line 19: 15 minutes seems generous, but I suppose it's better than the current situation where it waits indefinitely!

bormanp: I made it public so it can be turned into a flag :-)

We have tests that fail if they take more than an hour.  I decide 15 minutes was sufficient to a) it really probably will never succeed and b) short enough that it won't accidentally timeout a test.

bormanp:
I am not sure why lint is complaining about an empty line.  The only files whose last line is empty or consists solely of whitespace are:

```
./topo/node/cisco/testdata/push_config_success
./cloudbuild/vendors/topology.textproto
./cloudbuild/vendors/testbed.textproto
./cloudbuild/vendors/juniper.cfg
./examples/arista/ceos-150/arista.cfg
./examples/cisco/xrd/r2.config
./examples/cisco/xrd/r1.config
./examples/cisco/8000e/8000e.pb.txt
./examples/cisco/xrd-traffic/r2.config
./examples/cisco/xrd-traffic/r1.config
./examples/nokia/srlinux-services/2node-srl-ixr6-with-oc-services.pbtxt
./examples/nokia/srlinux-cert/2node-srl-with-cert.pbtxt
./examples/juniper/cptx-ixia/cptx-ixia.pb.txt
./manifests/base/kustomization.yaml
./.mk/kind.mk
```

None of those are Go files.

bormanp:
The presubmit seems to be failing in `TestMetalLBSpec` which is unrelated to this change.

alexmasi (approved):

PR #559, opened 2024-08-17
Add inside_ip to node and populate it in topo.Show

will2dye4:
This is needed for KNE clients to be able to configure the node's cluster-internal IP on the management interface for vendors that do not automatically configure the management interface with the node's IP.


proto/topo.proto
```diff
@@ -87,6 +87,8 @@ message Node {
   // Any host constraints required by the node. These may not be required by a
   // vendor, but it can be used to provide requirements for the node.
   repeated HostConstraint host_constraints = 13;
+  // Cluster IP for the node's pod. Assigned by KNE.
```
marcushines, line 90: I would suggest: pod_ip

as that is more accurate

IP assigned by k8s for the pod this ip comes from the worker node's IP pool

will2dye4: I chose `inside_ip` and phrased the comment to mirror the existing field on the `Service` proto message:

```
  string inside_ip = 4;   // Cluster IP for the service.
```

I think it's important to convey, either in the name or in comments or both, that this is an *internal* IP address, i.e., this IP cannot be used to reach the node outside of the cluster.

marcushines: while that "sounds good" inside ip and pod ip are actually different constructs

one ip is node worker specific another a common ip in the whole cluster that is used by the loadbalancer they are in fact different ip's and cluster inside ip means something 

as well as pod ip

will2dye4: That makes sense. I changed the name accordingly.

marcushines (approved):

PR #565, opened 2024-09-11
Keep kernel.pid_max smaller than 1048575.

guoshiuan:


topo/topo.go
```diff
@@ -484,6 +488,14 @@ func (m *Manager) load() error {
 		uid++
 	}
 	for k, n := range nMap {
+		// If the script is found, then run it. Else silently ignore it.
```
bormanp, line 491: This only needs to be done once per worker node, not per KNE node.

Isn't the set_pid_max.sh script called earlier than this?

guoshiuan: In the phase of "kne deploy ...", this value got changed. I also found some vendor changed this value during the topology creation. This is why I added it here.

bormanp: Change this comment to a BUG: and remove the reference to any particular vendor, just say this is addressing the issue that some vendors incorrectly lower this value which can cause other vendors to have issues.

bormanp (changes requested):

topo/topo.go
guoshiuan: done

bormanp (approved):

PR #570, opened 2024-11-04
Update kind to 1.24.0 and node to 1.31.0

marcushines:
also updates kube apis to 1.31


DanG100:
It doesn't look you changed in the kindest/node image anywhere? https://github.com/openconfig/kne/blob/main/deploy/kne/kind-bridge.yaml#L10

DanG100 (approved):

DanG100:
lgtm assuming you can fix presubmit

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openconfig/kne
 
-go 1.21
+go 1.22.0
```
bstoll, line 3: There are Go 1.21 version references in:

cloudbuild/external.pkr.hcl
cloudbuild/internal.pkr.hcl
.github/workflows/go.yml
docs/setup.md
x/wire/forward/Dockerfile

We should use the same go version in all of these (and use 1.23.2?)

PR #571, opened 2024-11-06
update packer

marcushines:


bstoll (approved):

DanG100 (approved):

cloudbuild/external.pkr.hcl
```diff
@@ -77,6 +77,10 @@ build {
       "echo \"fs.inotify.max_user_watches=25600000\" | sudo tee -a /etc/sysctl.conf", # configure inotify for cisco xrd containers
       "echo \"fs.inotify.max_queued_events=13107200\" | sudo tee -a /etc/sysctl.conf", # configure inotify for cisco xrd containers
       "echo \"kernel.pid_max=1048575\" | sudo tee -a /etc/sysctl.conf",              # configure pid_max for cisco 8000e containers
+      "sudo modprobe br_netfilter",
+      "sudo echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables",
```
bstoll, line 81: I missed this earlier - does this actually work? you usually need to do `echo 1 | sudo tee /proc/...`.  sudo echo redirects output as the user into procfs which won't work.

PR #578, opened 2024-11-20
Alpine: adding console node.

parth-opensrc:
For Alpine topology, this enables SSH to the host node through a service running on port 2222.


parth-opensrc:
Not needed now.

PR #611, opened 2025-04-19
Increase grpc MaxMessageSize limits from 4MB to 2GB.

pranaysameer:
When attempting to bring up topologies with large configs we hit message size limits. This change will increase limit from 4MB to 2GB


controller/server/main.go
```diff
@@ -630,6 +630,8 @@ func main() {
 			PermitWithoutStream: true,
 			MinTime:             time.Second * 10,
 		}),
+		grpc.MaxRecvMsgSize(math.MaxInt32),
```
bormanp, line 633: That value is an odd number.  It probably makes more sense to say 1 << 9.

bormanp (changes requested):

controller/server/main.go
pranaysameer: Done. Changed it to 1<<31.

bormanp (approved):

bormanp (approved):

PR #624, opened 2025-06-09
enable parallelization of pod creation

NehaManjunath:


topo/topo.go
```diff
@@ -598,23 +614,55 @@ func (m *Manager) push(ctx context.Context) error {
...
+			if err := nodeToCreate.Create(ctx); err != nil {
+				errCh <- fmt.Errorf("failed to create node %s: %w", nodeToCreate, err)
+				return
+			}
+			log.Infof("Node %s resource created", nodeToCreate)
+		}(n)
 	}
+	wg.Wait()
+
+	select {
```
marcushines, line 636: why not just send either nil or error down the channel then you don't need to select

topo/topo.go
```diff
@@ -598,23 +614,55 @@ func (m *Manager) push(ctx context.Context) error {
...
+	case err := <-errCh:
+		return err
+	default:
+	}
+
+	log.Infof("Generating Self-Signed Certificates")
```
marcushines, line 642: why not move the self signed gen as part of the previous func then you don't have two fan out / fan in

topo/topo.go
NehaManjunath: Done.

topo/topo.go
```diff
@@ -570,10 +571,25 @@ func (m *Manager) topologySpecs(ctx context.Context) ([]*topologyv1.Topology, er
...
 func (m *Manager) push(ctx context.Context) error {
+	var wg sync.WaitGroup
+	errCh := make(chan error, len(m.nodes)*2+3) // Buffer the error channel to avoid deadlocks
+
+	log.Infof("Validating Node Constraints")
 	for _, n := range m.nodes {
-		if err := n.ValidateConstraints(); err != nil {
-			return fmt.Errorf("failed to validate node %s: %w", n, err)
-		}
+		wg.Add(1)
```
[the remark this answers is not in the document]
NehaManjunath: As discussed offline, to ensure the behavior still matches the current order of operations post the parallelization , keeping validate separate. Failure in validation wouldn't proceed to namespace creation or any other operations.

marcushines (approved):

PR #625, opened 2025-06-12
Export default node specs

NehaManjunath:


topo/node/alpine/alpine.go
```diff
@@ -20,13 +20,31 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	log "k8s.io/klog/v2"
-	"k8s.io/utils/pointer"
+
+	apb "github.com/openconfig/kne/proto/alpine"
+	tpb "github.com/openconfig/kne/proto/topo"
+)
+
+var (
+	defaultNode = tpb.Node{
```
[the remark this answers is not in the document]
chrisy: > This vendor type doesn't have constraints defined. They have a todo item here: https://github.com/openconfig/kne/blob/main/topo/node/alpine/alpine.go#L273

I think it would be reasonable to come up with some defaults. I'd probably go with 0.5cpu core and 0.5gb ram as a placeholder for basic scripting and whatnot.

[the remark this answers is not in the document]
marcushines: +1 - i think let's just be arbitrary to move forward for now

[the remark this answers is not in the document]
NehaManjunath: Sounds good. Addressed comments.

topo/node/nokia/nokia.go
```diff
@@ -32,6 +33,9 @@ const (
 	// configuration reset is therefore done by reverting to this checkpoint
 	configResetCmd = "/tools system configuration checkpoint initial revert"
 	pushCfgFile    = "/home/admin/kne-push-config"
+
+	defaultCPU = "2"
```
marcushines, line 37: so at this point ...

defaultConstraints = node.Constraints{
   CPU: 2.0,
   Memory: 4 * 1 << 30, // gigabyte 
}

marcushines: then it is much clearer everwhere else

topo/node/arista/arista.go
```diff
@@ -44,6 +45,9 @@ import (
 
 const (
 	scrapliPlatformName = "arista_eos"
+
+	defaultCPU = "500m"
```
marcushines, line 49: this should be a standard pattern through here

and then you don't need "defaultCPU" "defaultMem" as seperate vars

NehaManjunath: Sounds good, addressed comments.

marcushines (approved):

PR #636, opened 2025-07-08
fix redundant steps in external pkr file

NehaManjunath:


cloudbuild/external.pkr.hcl
```diff
@@ -121,11 +121,6 @@ build {
       "echo Installing multinode cluster dependencies...",
       "git clone https://github.com/flannel-io/flannel.git",
       "curl --create-dirs -o third_party/licenses/flannel/LICENSE https://raw.githubusercontent.com/flannel-io/flannel/master/LICENSE",
-      "sudo modprobe br_netfilter",
```
jasdeep-hundal, line 124: In internal.pkr.hcl, we are doing it in the multinode section (https://github.com/openconfig/kne/blob/main/cloudbuild/internal.pkr.hcl#L119), so I'd prefer to do the same here so it's less likely we miss changes when modifying both.

jasdeep-hundal (changes requested):

cloudbuild/external.pkr.hcl
NehaManjunath: Agreed. This is fixed now.

jasdeep-hundal (approved):

PR #647, opened 2025-09-15
[Juniper] Only set the legacy gRPC services config if the specific node label isn't set.

gmacf:
These config lines are now not required in D47 and will cause future config commits to fail if combined with the newer gRPC services configuration.

From Juniper's D47 release notes:

`
NGINX-based server implementation allows for hosting GRPC services on ports specified in the system configuration. Configuration model for this feature is mutually exclusive from  the existing GRPC server configuration model. The new standards-based server implementation brings in performance enhancements for scaling gRPC sessions and has built-in support for the gNSI security suite. Support for using PKID-based certificate intake as well hot-reloading of TLS certificates is already built-in to the solution and does not need to be explicitly configured. 
`

`
This configuration should be present in the bootstrap configuration, and the “set system services extension-service request-response grpc” hierarchy must NOT be present while this is present, otherwise the system will fail to apply the configuration since the two models are mutually exclusive.
`

KNE topologies will be able to disable this legacy gRPC config by using this label.

Also incorporated a change from https://github.com/openconfig/kne/pull/622 that improved retry behavior on config pushes.

### DIFF
--- a/topo/node/juniper/juniper.go
+++ b/topo/node/juniper/juniper.go
@@ -33,11 +33,11 @@ var (
 	// For committing a very large config
 	scrapliOperationTimeout = 300 * time.Second
 	// Wait for PKI cert infra
-	certGenTimeout = 10 * time.Minute
+	certGenTimeout = 15 * time.Minute
 	// Time between polls
 	certGenRetrySleep = 30 * time.Second
 	// Wait for config mode
-	configModeTimeout = 10 * time.Minute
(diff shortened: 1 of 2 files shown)


gmacf:
Hi folks, could I get a review of this PR? Also need some help to re-run the presubmit after pushing https://github.com/openconfig/kne/pull/647/commits/107d1891e66091d1d8fb0d903ffd0c3e96fa39fe.

NehaManjunath (approved):

PR #653, opened 2025-10-15
Populate the correct non-legacy gRPC server config

gmacf:
In my previous PR  (https://github.com/openconfig/kne/pull/647), I removed the legacy gRPC service config as it was not compatible with D47. It turns out this is incorrect and left the device in a weird state.

With this PR, we ensure the gRPC server is configured fully for D47 instead of leaving it in a partially configured state.


bstoll:
Closing because I believe this is stale now.  The configs pushed look different since #656.

PR #686, opened 2026-03-23
Configure containerd mirrors for registry.k8s.io

NehaManjunath:


jasdeep-hundal (commented):
Not sure that the mirroring approach will work for us because I don't think there's a way to configure an additional path substitution `kne-external/kne/` (we can only use the actual host portion of the intended mirror).

jasdeep-hundal (approved):

PR #708, opened 2026-06-25
Add initial support for Ciena (saos and waverouter)

marcushines:
* Modified proto/topo/topo.proto
  * add Ciena vendor and ciena_saos type
* Modified proto/topo/topo.pb.go
  * changes generated by topo.proto
* Modified topo/topo.go
  * add ciena node into import
* Added topo/node/ciena/ciena.go
  * ciena node implementation
* Added proto/ciena.proto
  * add Ciena proto - define Ciena specific vendor data for KNE
* Added proto/ciena/ciena.pb.go
  * file generated by proto/ciena.proto


dpelton-ciena:
@marcushines - thanks for rebasing these changes.  If you need any additional input from Ciena folks, please let me know.

topo/node/ciena/ciena.go
```diff
@@ -0,0 +1,340 @@
...
+// renameInterfaces renames the interfaces port naming to ethX format
+// e.g., eth1 -> eth1, 1/5/1 -> eth1, 2 -> eth2, etc
+func renameInterfaces(interfaces map[string]*tpb.Interface) map[string]*tpb.Interface {
+    intf := map[string]*tpb.Interface{}
+    reFull := regexp.MustCompile(`^\d+/\d+/\d+$`)
+    reSingle := regexp.MustCompile(`^\d+$`)
+    for k, v := range interfaces {
+        switch {
+        case reFull.MatchString(k):
+            parts := regexp.MustCompile(`/`).Split(k, -1)
```
NehaManjunath: Nit: Recommend strings.Split(k, "/") over regex in the loop. 

Also, taking lastNum maps ports across different slots (e.g. 1/5/1 & 1/6/1) to the same eth1—wondering if we should incorporate slot numbers to avoid collisions and add unit tests for this case (is this is a valid scenario)?

topo/node/ciena/ciena.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		equipmentFile := cienaConfig.GetSystemEquipment().GetEquipmentJson()
+
+		contents, err := os.ReadFile(filepath.Join(n.BasePath, equipmentFile))
+		if err != nil {
+			return err
+		}
+		cm.Data[fileName] = string(contents)
+		extraMounts = append(extraMounts, corev1.VolumeMount{
+			Name:      "equipment-file",
```
NehaManjunath, line 200: Delete method in default node implementation [in node.go:565](https://github.com/openconfig/kne/blob/main/topo/node/node.go#L565), DeleteConfig filters specifically for vol.Name == ConfigVolumeName ("startup-config-volume"). 

Since the equipment file volume here is named "equipment-file", DeleteConfig skips it during kne delete. Would it make sense to add a custom Delete method to also clean up the <node>-file ConfigMap?

topo/node/ciena/ciena.go
```diff
@@ -0,0 +1,340 @@
+// Copyright 2025 Ciena Corporation
```
bstoll, line 1: cienna.go / cienna_test.go both need `gofmt` run on them (I will fix CI checks for this... I thought this was in there but it seems like it may not be.).

topo/node/ciena/ciena.go
dpelton-ciena: You are correct that the use of lastNum clobbers interfaces across different slots.  This code was written and tested using only a single interface card (i.e. only simulating one slot).  We (Ciena) will need to add more logic to properly handle the name to port allocation for multiple slots, as the number of interfaces per slot can vary depending on the card type.

To address this for now, we can either live with the one slot restriction or remove the buggy code that maps the "housing/slot/port" string and force the use of the equivalent "ethX" numbering.  Our internal KNE testing has been focused on the pizza box devices rather than the housing based devices.

NehaManjunath: Thank you, for this PR - we can probably just strip out the housing/slot/port mapping and force ethX. It would be nice to have to throw an error/warning for syntax on multi-slot.

We can perhaps handle the proper multi-slot allocation fix in a separate PR. WDYT?

dpelton-ciena: I agree that we will need to add a follow-up PR to add proper multi-slot support.

bstoll (approved):

PR #709, opened 2026-07-01
Import meshnet under KNE maintenance

kraney:
This takes the meshnet repo from networkop/meshnet-cni and imports it under a third_party directory to be maintained and released as part of the KNE project. It's placed there for specific reasons

* It's not really a go-style vendor/ directory since the original project is archived. This will be modified and maintained under openconfig/KNE control.
* It has a BSD license distinct from KNE's license. The third_party directory is Google's preferred way to maintain that separation.

This import keeps meshnet-cni largely unchanged. There are a few exceptions.
 * The package paths are updated to the new location to avoid linking into the original home
 * Because meshnet depends on protoc latest, the golang version specified in go.mod has been bumped up to meet protoc's current requirements so that meshnet will build
 * Go code for the protos has been regenerated from the current protoc
 * The version of Go used to build has been updated to meet the resulting requirements
 * The Docker base image has been updated to match the images used to build
 * Meshnet's github actions have been folded into KNE's

All of these changes are intended to make the minimal set of changes that keeps meshnet from being broken in its new home - no substantive functional changes. Improvements to meshnet will happen later in separate PRs.

Notably this PR does NOT alter KNE to make use of meshnet from its new home. The intent is that this PR will make it possible to release meshnet to a public-facing docker repo. Then, once that's available, KNE can be updated to rely on it and on this code as a separate PR. Only after that's in place will any more meaningful changes happen for meshnet.


chrisy:
I certainly approve of making meshnet more alive than it was; I've been maintaining a local copy myself, with some fixes in it. Curious why a subdir rather than its own repo?

kraney:
Within the team here, the sense is that it won't have other uses within openconfig outside of KNE, so there was a preference to avoid having another repo with its own maintenance and release cycle. Folded into KNE, it will become just part of the established workflow, team ownership, PR reviews, etc. It's subjective, but this was viewed as the simpler path.

bstoll (approved):

NehaManjunath (approved):

PR #712, opened 2026-07-09
Add cloudbuild config to release meshnet

kraney:
This builds a docker image for meshnet and pushes it to artifact registry. The approach is modeled after the approach used by lemming.

note, when this is merged, I recommend tagging the new revision as v0.3.3.
* without the tag, the resulting docker image will be labeled v0.3.1, from KNE’s last published release
* KNE currently uses a meshnet image labelled v0.3.2, so this way the image will be strictly greater than that
* KNE will skip v0.3.2, but that’s fairly trivial and this way meshnet and KNE will begin to have synchronized versioning.

I intend to follow fairly quickly with a pull request that will have KNE make use of its own internal meshnet rather than the networkops one. But I think it’s preferable for a useable docker image to already be in place before I make KNE start linking against the code associated with it.


cmd/release/release.go
```diff
@@ -0,0 +1,183 @@
...
+		return "", err
+	}
+	branch := strings.TrimSpace(string(brOut))
+	revOut, err := exec.Command("git", "rev-parse", "HEAD").CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	sha := strings.TrimSpace(string(revOut))
+	ready := true
+	if branch != "main" && branch != "master" {
```
marcushines: "master" should not be a thing at this point - we shoudl fix the branch rather than supporting this branch name

cmd/release/release.go
```diff
@@ -0,0 +1,183 @@
...
+		fmt.Println("Not on main or master branch")
+		ready = false
+	}
+	if status != "" {
+		fmt.Println("Working directory dirty")
+		ready = false
+	}
+	if !ready {
```
marcushines: i don't like this pattern as it is mixing io with what should just be a processing data

A better construct. would be to return to either provide in input to this function to say it is okay to skip these checks or you hande the error on return

cmd/release/main.go
```diff
@@ -0,0 +1,28 @@
+// Copyright 2024 Google LLC
```
marcushines: main should be the only thing in main

the rest of this should be a new package

cmd/release/release.go
```diff
@@ -0,0 +1,198 @@
+// Copyright 2024 Google LLC
```
marcushines, line 1: this should be a new package normally we make a cmds package which holds the cobra command tree

and in this case there would be release directory with this package in it

marcushines (approved):

PR #716, opened 2026-07-15
Move KNE to link to its own meshnet

kraney:
This updates package references to point to the meshnet now under third_party/meshnet, and updates docker
image references to point to us-west1-docker.pkg.dev/kne-external/kne/meshnet:ga rather than networkops/meshnet-cni.

This (once tagged and released) completes the process of bringing meshnet under KNE maintenance and will allow future enhancements to meshnet.


manifests/meshnet/grpc/manifest.yaml
```diff
@@ -272,46 +282,46 @@ spec:
...
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          image: us-west1-docker.pkg.dev/kne-external/kne/meshnet:ga
+          imagePullPolicy: IfNotPresent
```
chrisy, line 307: Writing this more as a note-to-self, though it may be worth a note in a README somewhere: Persistent kube setups will need the cached image to be purged to see new versions of meshnet; `:ga` will be sticky otherwise.

bstoll (approved):
sorry about the mess of formatting changes you had to go through, I was trying to independently fix those up so you wouldn't have to go through that mess.

PR #717, opened 2026-07-16
Remove limitation preventing high link counts

kraney:
Meshnet has had a practical limitation on the number of links any pod can have as well as the aggregate number of links on any K8s node.
* Links were set up within the CNI plugin
* The CNI plugin has a hard time limit generally around 90 seconds
* The process requires coordination between both ends of a link, which involves an API call via meshnetd to kube-apiserver and those calls are rate limited

The net effect is that it's hard to exceed around 250 interfaces in a KNE node. What's more the failure mode is not graceful, when the CNI times out K8s tries to destroy the container sandbox and start over, leading to a loop it can't escape.

This change removes this limitation using the following approach.
* The work of adding interfaces to a pod is removed from the CNI plugin. It now just publishes its presence to meshnetd with SetActive(), which provides necessary information including its network namespace ID.
* Meshnetd acts as a typical K8s controller, watching for changes to the topology. Once both ends report themselves active, the local meshnetd on the k8snode does the work of preparing interfaces for the container.
* Because meshnetd handles both ends of the link itself, the coordination between pods becomes unnecessary reducing the need to talk to kube-apiserver to check state. For links between nodes, each meshnetd can work independently without coordination.
* For vxlan, use an offset for the ID block based on namespace so we don't fail mysteriously from collisions if more than one topology is present.
* If there are errors, they're reported in the topology status, not just logged, for easier detection and debugging.
* The setup process uses deterministic naming for the resource created so that it's idempotent; if meshnetd happens to crash or get restarted, it can resume the process where it left off.

Also, the CNI plugin now waits up to 15 seconds after SetActive() for expected links to be present. We set a maximum wait to stay well clear of K8s' time limit and the chaos that ensues if it's exceeded by the CNI plugin. But we give enough time that for the common case, and most if not all existing cases, the CNI plugin doesn't return until all interfaces are present. This avoids the need to add an init-wait init container to pods for simple tests with low link count. Only when the work gets significant, init-wait becomes necessary. Note that this is a small improvement over the prior behavior, since one of the two pods for any link would skip creating that interface, and thus wouldn't wait for that particular one to be present.

Tests that force pod affiinity (same node veth links) or pod antiaffinity (separate node vxlan or grpcwire links) have been added to validate that the problem is addressed.


chrisy:
> * The work of adding interfaces to a pod is removed from the CNI plugin. It now just publishes its presence to meshnetd with SetActive(), which provides necessary information including its network namespace ID.
> * Meshnetd acts as a typical K8s controller, watching for changes to the topology. Once both ends report themselves active, the local meshnetd on the k8snode does the work of preparing interfaces for the container.
> * Because meshnetd handles both ends of the link itself, the coordination between pods becomes unnecessary reducing the need to talk to kube-apiserver to check state. For links between nodes, each meshnetd can work independently without coordination.
> ...

Thank you for removing an item off my backlog! A nice side effect is this also cures a race condition; it's easily worked around, but on fast kube nodes, it's possible for both ends of a link on the same server to think they should be creating the veth pair. One fails because it exists already and it never recovers. The workaround is to ignore certain errors, but it's a kludge (there was a report of this, never resolved, on the original meshnet repo).

kraney:
Actually there's another nice side effect in that the CNI-based algorithm could only work at initialization time, but the new meshnetd one can potentially react to topology changes later in the lifecycle. I didn't really test that use case since it didn't work before, so there might be a bit more work to resolve issues with dynamic changes before it can really be claimed to work. But the potential is there.

chrisy:
> Actually there's another nice side effect in that the CNI-based algorithm could only work at initialization time, but the new meshnetd one can potentially react to topology changes later in the lifecycle. I didn't really test that use case since it didn't work before, so there might be a bit more work to resolve issues with dynamic changes before it can really be claimed to work. But the potential is there.

That was not lost on me, but I didn't want to get ahead! Though many vendor containers only enumerate their ports at startup, some do not, and perhaps could be worked around with placeholder veth's. Tangentially, I think it also simplifies something I've wanted to look at around managing link state and applying Linux TC rules, which I suspect requires us to always wire veth's via meshnetd (instead of sometimes direct when pods are on the same node.)

kraney:
> > Actually there's another nice side effect in that the CNI-based algorithm could only work at initialization time, but the new meshnetd one can potentially react to topology changes later in the lifecycle. I didn't really test that use case since it didn't work before, so there might be a bit more work to resolve issues with dynamic changes before it can really be claimed to work. But the potential is there.
> 
> That was not lost on me, but I didn't want to get ahead! Though many vendor containers only enumerate their ports at startup, some do not, and perhaps could be worked around with placeholder veth's. Tangentially, I think it also simplifies something I've wanted to look at around managing link state and applying Linux TC rules, which I suspect requires us to always wire veth's via meshnetd (instead of sometimes direct when pods are on the same node.)

Yes ideally if we're going to move links around, we should leave the port the pod is using where it is, and just change what it's linked to. Which isn't doable with a veth pair, but might be doable with e.g. an OVS port in the container, just changing the forwarding rules. That way the change is more like moving a fiber cable, and less like removing the entire NIC and putting it back which is quite weird from a hardware vendor perspective.

I'm doubtful linux native VXLAN can do that. I think the VTEP is immutable after netdev creation (might be wrong.)

Grpcwire could since it's just using libpcap. However I'm very concerned that using libpcap creates a high risk of packet loss when the TX is faster than grpcwire can transmit. There's no path for backpressure, libpcap has a core requirement not to delay the "real" tx, it's intended to be an optional side channel. Also, grpcwire copies the packet out of kernel space, encodes it into the protobuf, then copies that into kernel space for transmit. Then does it all in reverse on the other side. 6 full copies per packet. In the current implementation it looks like it makes a synchronous RPC, so it waits for the rpc round trip before starting on the next packet. All that adds up to it very likely falling behind with resulting packet drops.

marcushines:
ping on fix conflict

marcushines (approved):

PR #719, opened 2026-07-17
Update juniper to restart session.

kraney:
There's a potential race when there's a schema change, that leaves a session broken. In this case, Juniper outputs "Restart session and try again."

The existing code tries again, but without restarting the session. So it is doomed to just fail again. This change notices the condition and restarts the session if it occurs. In theory that should help.

This is intended to unblock PR #717, so may need to merge in front of that one.


kraney:
Withdrawing this because it did not, in fact, improve the result.

PR #725, opened 2026-07-23
build(cloudbuild): use pd-ssd disk type for VM creation and packer builds

bstoll:
Use pd-ssd disk type for faster builds.


bstoll:
With regular HDD disk presubmit takes about 33m, with this change its about 18m.

kraney (approved):
lgtm

PR #729, opened 2026-07-29
fix(node): sanitize special characters and escape sequences in config

bstoll:
- Add character allowlist validation for Juniper certificate names in waitCertInfraReadyAndPushCert and GenerateSelfSigned.
- Add character allowlist validation for Cisco XRd startup configuration paths in ResetCfg.
- Improve escaping of special characters (\, ", `, and $) in Nokia SR Linux configuration strings before sending via CLI.


kraney (approved):
lgtm

PR #731, opened 2026-07-31
[meshnet] Replace veth + libpcap with TAP devices for gRPC wires

kraney:
Replace the veth pair and libpcap packet capture implementation in meshnetd
    with persistent Linux TAP devices created directly inside container network
    namespaces.

    Reason for Change:
    - Eliminates packet loss under high traffic loads caused by libpcap's kernel-
      to-userspace drop policy.
    - Provides native Linux socket buffer flow control and backpressure (`txqueuelen`)
      up to the container application when gRPC processing falls behind.
    - Guarantees protocol transparent packet handling (including LACP and LLDP frames).
    - Ensures process crash resilience: setting `TUNSETPERSIST=1` keeps TAP interfaces
      alive in the container netns across daemon restarts without link flaps.
    - Synchronizes CNI plugin pod readiness with complete end-to-end gRPC wire setup,
      preventing test/ping race conditions upon pod startup.
    - Eliminates CGO and `libpcap-dev` build dependencies, producing a pure Go static
      binary (`CGO_ENABLED=0`).

    Key Changes:
    - Added `CreateOrAttachTAP` in `wireutil` using `TUNSETIFF` & `TUNSETPERSIST`.
    - Replaced `pcap.Handle` with `*os.File` in `gwire_map.go`, `grpcwire.go`, and `handler.go`.
    - Updated `ReconcilePodLinks` in `controller.go` to use TAP interfaces directly without
      host-side veth creation.
    - Updated `GRPCWireExists` and CNI `cmdAdd` readiness check to block until gRPC wire
      handshakes are fully established.
    - Removed `libpcap-dev` and updated Dockerfile to build `meshnetd` with `CGO_ENABLED=0`.


kraney:
FWIW, I created a throughput test comparing before & after for grpcwire. ("Before" being before #731 , "after" being after #734 , including the full set of improvements.)

Before:
image

After:
image

chrisy:
> FWIW, I created a throughput test comparing before & after for grpcwire. ("Before" being before #731 , "after" being after #734 , including the full set of improvements.)

Well, that is certainly compelling!

third_party/meshnet/daemon/grpcwire/grpcwire.go
```diff
@@ -344,45 +304,67 @@ func RecvFrmLocalPodThread(wire *GRPCWire, locIfNm string) error {
...
-	var packet gopacket.Packet
+	buf := make([]byte, 65535)
+	type readResult struct {
+		n   int
+		err error
+	}
+	readChan := make(chan readResult, 1)
+	go func() {
+		for {
+			n, err := tapFile.Read(buf)
```
bstoll: The contents of buf can be overridden by a 2nd read before the first is copied (data corruption on wire). This is also a hot code path - its best to not allocate memory on every message because this creates a lot of pressure on the garbage collector (line 346).

Can a test be written for this?

third_party/meshnet/plugin/meshnet.go
```diff
@@ -208,32 +208,52 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-					log.Infof("Add[%s]: Bounded readiness wait expired (%d links); asynchronous completion will continue", string(cniArgs.K8S_POD_NAME), len(localPod.Links))
-					return nil
-				case <-ticker.C:
+
+				// For gRPC links, check if the gRPC wire is fully established on the daemon
+				wireDef := &mpb.WireDef{
+					LocalPodNetNs: args.Netns,
+					LinkUid:       link.Uid,
+				}
+				resp, err := meshnetClient.GRPCWireExists(waitCtx, wireDef)
```
bstoll: Maybe this is an issue for another PR, but this section seems flawed to me.  This is waiting 30 seconds for interfaces to come into existence, but if the 30 seconds passes it continues anyway. Either we didn't need to wait at all, or this is leaving things in a broken state and continuing anyway after 30 seconds.

Should this check that we are actually using GRPC wires here? I think this loops over even local interfaces, which will never return true on meshnetClient.GRPCWireExists, which means pods must wait 30 seconds for something to never happen.

third_party/meshnet/daemon/meshnet/controller.go
```diff
@@ -171,70 +168,28 @@ func (m *Meshnet) reconcilePodLinksInternal(ctx context.Context, topo *unstructu
...
-					PeerNodeIp:            peerSrcIP,
+					LocalPodNetNs: netNS,
+					LinkUid:       link.LinkUID,
+					TopoNs:        link.KubeNs,
+					LocalPodName:  topo.GetName(),
+					IntfNameInPod: link.LocalIntf,
+					LocalPodIp:    link.LocalIP,
+					PeerNodeIp:    peerSrcIP,
 				}
 				if _, err := grpcwire.CreateGRPCWireLocal(ctx, wireDefLocal); err != nil {
```
bstoll, line 181: AI comment:

  • Location: controller.go and grpcwire.go
  • Mechanism: In controller.go:181, CreateGRPCWireLocal starts RecvFrmLocalPodThread listening on aWire.StopC. At line 223, reconcilePodLinksInternal calls:
    grpcwire.UpdateWireByUID(netNS, int(link.LinkUID), creatResp.PeerIntfId, make(chan struct{}))
    UpdateWireByUID overwrites wire.StopC with a brand new channel without holding wire.mu.
  • Impact: When WireDownByUID or RemoveWireAcrosAll closes wire.StopC, it closes the new channel, while RecvFrmLocalPodThread is still listening on the old channel. The receive thread never receives the stop signal and runs indefinitely.
  • Fix: Do not recreate or overwrite wire.StopC in UpdateWireByUID if it is already initialized; hold wire.mu when modifying wire state

third_party/meshnet/utils/wireutil/tap.go
```diff
@@ -0,0 +1,82 @@
...
+		ifr.Flags = unix.IFF_TAP | unix.IFF_NO_PI
+
+		_, _, errno := unix.Syscall(unix.SYS_IOCTL, uintptr(fd), uintptr(unix.TUNSETIFF), uintptr(unsafe.Pointer(&ifr)))
+		if errno != 0 {
+			unix.Close(fd)
+			return fmt.Errorf("TUNSETIFF failed for %s in netns %s: %v", ifName, podNsPath, errno)
+		}
+
+		// Make device persistent so it survives process crashes/restarts
+		_, _, _ = unix.Syscall(unix.SYS_IOCTL, uintptr(fd), uintptr(unix.TUNSETPERSIST), 1)
```
bstoll: nit: check errno here too

third_party/meshnet/utils/wireutil/tap.go
```diff
@@ -0,0 +1,82 @@
...
+			return fmt.Errorf("failed to find link %s inside netns %s: %w", ifName, podNsPath, err)
+		}
+
+		if err := netlink.LinkSetUp(link); err != nil {
+			log.Warnf("CreateOrAttachTAP: failed to set %s UP in netns %s: %v", ifName, podNsPath, err)
+		}
+
+		if ipCIDR != "" {
+			addr, err := netlink.ParseAddr(ipCIDR)
+			if err == nil {
```
bstoll: Why are errors being discarded here?

third_party/meshnet/plugin/meshnet.go
kraney: This function would complete successfully even without waiting at all. But the wait serves a purpose - in the common case, when all interfaces can be in place before the CNI timeout is a factor, the wait here means that the pod will not start the main container until all interfaces are present. That helps prevent the earlier change where interfaces are added separately from the CNI plugin from breaking existing cases that expect interfaces to be present.

at the same time, there’s a hard limit on how long it is possible to wait, about 90s. Beyond that, the pod startup will explode. So this makes a best effort, but gives up before that becomes a factor. Cases that need a longer wait must use the init-wait startup container.

init-wait has always been a practical requirement because before, pods only actually waited for about half of their interfaces, with the other end waiting on the other half. This new version waits for all up to the time bound.

kraney: Regarding veth, this function is not called in the veth case.

kraney: > Regarding veth, this function is not called in the veth case.

No, I'm mistaken. Will fix

third_party/meshnet/daemon/grpcwire/grpcwire.go
kraney: Excellent catch, thanks very much. Will include a fix and a test.

third_party/meshnet/daemon/meshnet/controller.go
kraney: Fixed with locking semantics

bstoll (approved):

PR #733, opened 2026-07-31
[meshnet] multiplex tcp streams for a node pair

kraney:
Note: this PR stacks on top of #732 , please merge that one first. The incremental changes start at sha [ceb9511](https://github.com/openconfig/kne/pull/733/commits/ceb951110d1e1efc3ecb3a30650676ae51c105fd)

      1. Created nodeStreamManager (stream_manager.go):
          • Keyed by nodeStreamKey{ topoNs string, peerIP string }.
          • Opens only one gRPC connection (grpc.Dial) and only one SendToStream 
             bidirectional streaming RPC per (peerIP, topoNs) pair.
          • Includes reference counting (GetOrCreateStream / ReleaseStream). When the last 
             wire for a topology targeting a peer node is removed, the shared stream and gRPC 
             connection close gracefully.
          • Features a high-capacity buffered channel (10,000 packet queue) and dedicated sender 
             worker loop with automatic reconnect logic.
      2. Updated TAP Reader Threads (grpcwire.go):
          • Modified RecvFrmLocalPodThread so TAP readers no longer execute individual grpc.Dial 
             or SendToStream calls.
          • Each TAP reader now acquires the shared topology stream via 
             nodeStream := streamMgr.GetOrCreateStream(wire.TopoNamespace, wire.PeerNodeIP) 
            and multiplexes packets into nodeStream.Send(payload).
      3. Per-Topology Isolation:
          • Topology topo-A and topology topo-B running between the same pair of physical 
              nodes maintain completely separate gRPC connections and streams.
          
Also:
    The base design had a lot of single-interface RPCs
    to update k8s and remote meshnets. For large numbers
    of links it costs a lot of serial round trip wait time.
    This batches things and pipelines them so that we avoid
    a lot of unnecessary overhead delay at start time.

PR #734, opened 2026-07-31
[meshnet] Manage system-level kernel and link tunables for packet performance

kraney:
Note: this PR stacks on top of #733 . Please merge that one first (which should make the shared commits disappear from this PR.) The incremental change of this PR starts with sha:b3c915457042d606dd0dd9bb659ed1ea619af81e.

        Implement kernel sysctl tuning in meshnetd for high-density, high-throughput network meshes.

        Reason for Change:
        - Default Linux link queue lengths, netdev backlogs, multicast membership limits, and
          strict rp_filter settings cause packet drops during netdev creation bursts and multi-link tests.
        - IPv6 DAD (Duplicate Address Detection) and Router Solicitations introduce 1-2s readiness
          delays per netdev when launching virtual topologies.

        Key Changes:
        - System & Kernel Tuning (`utils/wireutil/sys_tune.go`):
          - Sets `RLIMIT_NOFILE` to 1,048,576 open file descriptors.
          - Sets `netdev_max_backlog` to 10,000 for device queue bursts.
          - Sets OS send/recv max & default buffers (`rmem_max`, `wmem_max`, `rmem_default`, `wmem_default`) to 16 MB.
          - Sets ARP/neighbor GC thresholds (`gc_thresh1`=1024, `gc_thresh2`=4096, `gc_thresh3`=8192).
          - Sets `igmp_max_memberships` to 10,000 and `mld_max_msf` to 4,096 to prevent multicast group netdev drops.
          - Sets `rp_filter` to 2 (Loose mode) to allow asymmetric routing across multi-link meshes.
          - Disables IPv6 DAD (`accept_dad`=0) and RS floods (`router_solicitations`=0) for instant link readiness.
          - Sets `ipv6/route/max_size` to 1,048,576 for large routing tables.
          - Configured link `txqueuelen` (default 10,000) across TAP, vEth, and vxLAN interface creation.
          - All tunables are overridable via environment variables without recompilation.


kraney:
FWIW, I created a throughput test comparing before & after for grpcwire. ("Before" being before #731 , "after" being after #734 , including the full set of improvements.)

Before:
image

After:
image

third_party/meshnet/daemon/main.go
```diff
@@ -22,6 +22,8 @@ func main() {
 	}
 	defer cni.Cleanup()
 
+	wireutil.TuneSystem()
```
bstoll: Can we move all of this to documentation instead (explaining why we are changing these things?) I think these kind of settings are really up to the user to implement - we should not be forcing tuning settings in code like this. We could make changes to cloudbuild/internal.pkr.hcl and cloudbuild/external.pkr.hcl if there are changes that will improve performance measurably in the VM images.

I don't think many of these changes are relevant or helping: ARP/ND cache GC thresh doesn't do much, rp_filter is not a performance knob, ipv6/route/max_size should not be changed, default/router_solicitations , accept_dad, mld_max_msf, and igmp_max_memberships don't do much in the host OS and are not going to change performance.  RLIMIT_NOFILE seems like its changing something for a limit we've never hit?

The txqueuelen, netdev_max_backlog, and [wr]mem_(default|max) I can imagine are going to affect performance. I think setting the initial window size to 16MB is a little excessive, but increasing the max seems OK.

kraney: I may have gotten overzealous trying to clear the last packet drops. Let me revisit

kraney: This is much more focused and justifiable now.

bstoll (approved):

PR #736, opened 2026-08-07
Update the pod anti-affinity configuration

kraney:
Updates third_party/openconfig/kne/topo/node/node.go to match on the namespace instead of the topology name, ensuring correct pod scheduling.

Otherwise, anti-affinity doesn't actually match and isn't enforced.


bstoll:
This problem also exists in the vendor implementations - can we change them all?

ie,
https://github.com/openconfig/kne/blob/main/topo/node/cisco/cisco.go#L237

kraney:
You're right - repeated the change across vendors.

bstoll (approved):

PR #739, opened 2026-08-12
Allocate more nodeports

kraney:
DUTs set up nodeports for their service ports. That's a scarce resource; the default K8s config allocates range 30000-32767 which limits to 2768 services total. Lemming has two services, as an example, so that leaves a max of 1384 instances possible, if nothing else is using nodeports.

This boosts the range by starting at 10000, giving more room for more nodes in the KNE topology.


bstoll:
When a kubeadm cluster is created couldn't it pass a config so it creates in the correct state instead of starting, modifying, and restarting it?

ie something like:

https://github.com/openconfig/kne/blob/1414e0c176f0e10e52a606c7f887f337d222542e/deploy/deploy.go#L454
`kubeadm init [opts] --config clusterconfig.yml`
...
```
    apiVersion: kubeadm.k8s.io/v1beta3
    kind: ClusterConfiguration
    apiServer:
      extraArgs:
        service-node-port-range: 10000-32767
```

kraney:
> When a kubeadm cluster is created couldn't it pass a config so it creates in the correct state instead of starting, modifying, and restarting it?

It is a little funky, but has a reason:

  1. CLI Flags vs. Config API Versioning:
      • kubeadm CLI flags (--cri-socket, --pod-network-cidr, --image-repository) are stable across almost all Kubernetes releases.
      • kubeadm YAML config API schemas are versioned and tied to specific Kubernetes releases (e.g., v1beta3 for Kubernetes 1.22–1.30, v1beta4
      for 1.31+). If the installed kubeadm version doesn't match the apiVersion in the generated YAML, kubeadm init --config fails.
  2. Flag / Config Mutual Exclusivity:
      • In newer kubeadm versions, passing CLI flags alongside --config produces warnings or errors. Everything (CRISocket, PodNetworkCIDR,
      ImageRepository, TokenTTL, ServiceNodePortRange) must be rendered into the YAML document.
  3. Current Post-Init Manifest Editing:
      • kubeadm init generates /etc/kubernetes/manifests/kube-apiserver.yaml.
      • SetServiceNodePortRange modifies this static pod manifest. kubelet automatically detects file changes and reloads kube-apiserver.
      • Pros: Version-agnostic; relies only on standard CLI flags for kubeadm init.
      • Cons: kube-apiserver restarts once right after cluster creation.

kraney:
After discussion, moved to the config-based strategy.

bstoll (approved):

PR #740, opened 2026-08-13
Set up default route for management network for Juniper and Cisco

kraney:
Both Juniper and Cisco ignore the K8s-configured settings on eth0 in their container. They set up everything from their own static config, but that makes the management interface unreachable in a multi-node cluster. Setting up the correct routing for the management interface is difficult in a static config since it's different depending on which node the pod has been assigned to at runtime.

This change has KNE update the routing config for the container at runtime to put back the routing set by Kubernetes.


topo/node/juniper/juniper.go
```diff
@@ -297,6 +319,9 @@ func (n *Node) waitCertInfraReadyAndPushCert() error {
 
 // GenerateSelfSigned generates a self-signed TLS certificate using Junos PKI
 func (n *Node) GenerateSelfSigned(ctx context.Context) error {
+	if n.KubeClient == nil {
+		return errors.New("kubeclient is nil")
+	}
 	selfSigned := n.Proto.GetConfig().GetCert().GetSelfSigned()
 	if selfSigned == nil {
```
bstoll, line 308: If there is no cert to config this exits and avoids applying the route config -- similarly it's a little strange that the GenerateSelfSigned() function on Cisco doesn't generate a self signed certificate at all, it only does this route modification.

It seems like it'd be a lot cleaner if we had some kind of generic post-startup actions that could be executed and this would be a new type of action instead of trying to piggyback on an existing function.

Is there no way to figure out the gateway before the pod starts so this could just be injected in the configmap? I think this might be a lot simpler if these extra lines could be in the base config and you'd avoid this problem.

kraney: You're right, the naming is misleading. I'll revise for that.

The routing required isn't really known until the pod has been scheduled; it depends on which node it lands on. We could conceivably use an init container to update the configmap before letting the main pod start. However this has a downside - the init container needs the permissions to do it; either to make a request to KNE or the apiserver. I suppose another option would be for the init container to just hold until KNE confirms pre-startup preparations are complete.

kraney: This is reimplemented now as an init container function. That has two major benefits:
Since we're not racing with the DUT clearing out the network settings, we can just pick up the correct values from eth0. Not apiserver calls needed.
As mentioned, we just write the config into the startup configuration. It's in place from the beginning.

bstoll (approved):

PR #747, opened 2026-08-20
Add a K8s readiness check for device pods.

kraney:
Cisco pods may have a running process, but fail to come to "Router up" state. Prior to this change, KNE just hangs forever waiting for Router up even for permanent failures.

This resolves this in two ways (belt and suspenders.) First, for Cisco pods we add the check for "Router up" as a readiness check. That way, a failure ends up reflected in K8s' reported pod status. The pod will not transition to Ready state until that message is logged. And if it doesn't show up within the threshold, the pod will report as failed.

Second, KNE explicitly checks for failure in the logs and treats it as such, rather than just waiting indefinitely. KNE may notice first, since the failure message will likely arrive before the readiness check exceeds its threshold.

Similar weaknesses exist broadly across vendors - KNE typically waits for Running, not Ready. This means it waits for the internal process to start, not for it to be ready - and many of the device types take an appreciable time to actually be ready. The pods themselves appear to be running even if configuration has failed.

This sets up readiness checks broadly, so that startup failures are reported as startup failures and on a paricular single pod, rather than being silently ignored and letting the test fail at a later stage, often based on a timeout trying to connect to gnmi, or something similar.


bstoll:
The problem with requiring these readiness probes is they force users to always run these services - cisco / gnmi as an example requires the user to explicitly enable gnmi - the default configs do not do this.  What if someone configures gnmi on a different port, or doesn't configure it at all? The pods would never come up.

I don't know if this is the best method but one approach could be to do it dynamically based on if the topology config defined the service. If there is an "ssh" or "gnmi" service defined, we will create a readiness probe for one of them. This would let us create a readiness probe based on the port numbers the user defined for the service. It still assumes these specific services will also be enabled by the base/default config, but there is probably some value to having in readiness probe on pods - maybe we treat "gnmi" and "ssh" as special and will probe for them on startup if the topology config defined them?

kraney:
> The problem with requiring these readiness probes is they force users to always run these services - cisco / gnmi as an example requires the user to explicitly enable gnmi - the default configs do not do this. What if someone configures gnmi on a different port, or doesn't configure it at all? The pods would never come up.
> 
> I don't know if this is the best method but one approach could be to do it dynamically based on if the topology config defined the service. If there is an "ssh" or "gnmi" service defined, we will create a readiness probe for one of them. This would let us create a readiness probe based on the port numbers the user defined for the service. It still assumes these specific services will also be enabled by the base/default config, but there is probably some value to having in readiness probe on pods - maybe we treat "gnmi" and "ssh" as special and will probe for them on startup if the topology config defined them?

Yeah, valid point. I think dyanmism definitely helps; revised so that it picks up either ssh or gnmi service defs if possible and relies on them, or doesn't do the readiness check if it doesn't find them.

bstoll:
AI Review:

### 1. Cisco XRD Bypasses Readiness Probe Completely

In `topo/node/cisco/cisco.go`, `Create()` attaches `ReadinessProbe: node.ServiceReadinessProbe(pb)` to the container. But in `Status()`:

```go
case corev1.PodRunning:
    pb := n.Proto
    if pb.GetModel() != ModelXRD {
        // 8000e log checks
    }
    return node.StatusRunning, nil
```

When an XRD pod starts:
1. Kubernetes sets `Phase = corev1.PodRunning` within ~1–2 seconds.
2. Because XRD takes several minutes to boot control-plane processes, the readiness probe fails initially and Kubernetes reports `PodReady = False`.
3. In `Status()`, `pb.GetModel() != ModelXRD` is `false`, so it skips the `if` block and immediately returns `node.StatusRunning, nil`.

**Effect**: KNE declares XRD ready within 1–2 seconds, before the router is booted or listening on its ports, completely bypassing the readiness probe.

**Fix**: For XRD (or all Cisco models), inspect `p[0].Status.Conditions` for `corev1.PodReady == corev1.ConditionTrue` before returning `node.StatusRunning`:

```go
case corev1.PodRunning:
    pb := n.Proto
    if pb.GetModel() != ModelXRD {
        // 8000e checks
    }
    for _, cond := range p[0].Status.Conditions {
        if cond.Type == corev1.PodReady && cond.Status == corev1.ConditionTrue {
            return node.StatusRunning, nil
        }
    }
    return node.StatusPending, nil
```

---

### 2. Duplicate Pod Log Streaming in Cisco Status Polling

In `topo/node/cisco/cisco.go:630-645`:

```go
req := n.KubeClient.CoreV1().Pods(p[0].Namespace).GetLogs(p[0].Name, &corev1.PodLogOptions{})
if isNode8000eFailed(ctx, req) {
    ...
}
req = n.KubeClient.CoreV1().Pods(p[0].Namespace).GetLogs(p[0].Name, &corev1.PodLogOptions{})
if !isNode8000eUp(ctx, req) {
    ...
}
```

Every 500ms during `checkNodeStatus`, `GetLogs` is called **twice consecutively** for the same pod, streaming full logs twice over the network from `kube-apiserver`.

**Fix**: Fetch the log stream once into a buffer and run both regex checks against that single buffer:

```go
podLogs, err := req.Stream(ctx)
if err != nil {
    return node.StatusPending, nil
}
defer podLogs.Close()

buf := new(bytes.Buffer)
if _, err := io.Copy(buf, podLogs); err != nil || buf.Len() == 0 {
    return node.StatusPending, nil
}
logBytes := buf.Bytes()
if podIsFailedRegex.Match(logBytes) {
    return node.StatusFailed, fmt.Errorf("cisco %s node %s failed to boot", n.Proto.Model, n.Name())
}
if !podIsUpRegex.Match(logBytes) {
    return node.StatusPending, nil
}
```

*(Also note: check `p[0].Status.Conditions` for `PodReady` here as well so 8000e actually validates that the readiness probe passed).*

---

### 3. Missing Unit Tests

- **`topo/node/cisco/cisco_test.go`**:
  - Add tests for `isNode8000eFailed` matching fatal error strings (`Router failed to come up`, `FATAL sim:`, `LoginTimeoutError`).
  - Add test verifying `Status()` returns `node.StatusFailed` when the log contains failure patterns.
  - Update `TestCreate` to assert that `ReadinessProbe` is properly populated on the pod.
- **`topo/node/juniper/juniper_test.go`**:
  - Update `TestCreate` to verify `ReadinessProbe` on the created pod (consistent with tests added for Sonic and Ciena).

---

### 4. Nitpicks & Style

1. **Shadowing built-in `len`**:
   In `isNode8000eFailed` (and `isNode8000eUp`):
   ```go
   len, err := io.Copy(buf, podLogs)
   ```
   `len` shadows the Go built-in function. Suggest `n, err := io.Copy(...)`.

2. **Non-deterministic map iteration in `ServiceReadinessProbe`**:
   In `topo/node/node.go`:
   ```go
   for k, svc := range pb.Services {
   ```
   Iterating over a Go map is randomized. If a node defines multiple services matching the preferred names, probe port selection could vary across runs. Sorting keys prior to iteration ensures determinism.

3. **PR Description Alignment**:
   The PR description states readiness checks inspect logs for `"Router up"` and that failures cause K8s to mark pods as failed. In reality, the PR implements TCP socket probes on `ssh`/`gnmi` ports, and K8s readiness probe failures keep pods in `PodRunning` with `PodReady=False`. Updating the PR description will avoid confusion for future maintainers.

bstoll (approved):

PR #753, opened 2026-08-24
New Ixia-c operator and configmap

octpetre:
- New ixia-c operator and KS configmap
- Removed rbac-proxy for the ceos, srlinux and lemming controllers


octpetre:
wrong cla

PR #755, opened 2026-08-25
fix(deploy): fix flaky context cancellation tests in health waiters

bstoll:
#### Problem

In MeshnetSpec.Healthy and deploymentHealthy, the wait loop evaluates a select between case <-ctx.Done(): and case e, ok := <-ch:. When a pre-canceled context is passed while the watch channel already has queued events representing a healthy status (e.g. during
  unit tests using fakeWatch), Go's select statement selects between ready cases pseudo-randomly.

When the event channel case is selected over <-ctx.Done(), the function returns nil (healthy) instead of the context error, causing intermittent test failures expecting "context canceled".

[Example build failure](https://github.com/openconfig/kne/actions/runs/32785628482/job/97616846811) `FAIL: TestMetalLBSpec/canceled_ctx`

#### Reproduction Command

    go test -race -cpu=1,2,4,8 -count=1000 -run "TestMeshnetSpec/canceled_ctx|TestMetalLBSpec/canceled_ctx" ./deploy


kraney (approved):
lgtm

PR #757, opened 2026-08-25
feat(deploy): update k8s to 1.36 and add cluster version pinning/validation

bstoll:
This updates KNE's default Kubernetes version to v1.36.1 and Kind to v0.32.0.

- Cluster version validation: Verify the API server's major/minor version during `kne deploy` match expected versions.
- Kubeadm version pinning: Added `kubernetes_version` to KubeadmSpec and wired it through the controller proto so kubeadm deployments can declare which version they expect instead of inheriting whatever kubeadm binary happens to be installed on the host.
- Presubmit host syncing: Cloud Build runs on existing VM images. `kne_test.sh` and `vendors_test.sh` now check the host packages against `external.pkr.hcl` and update them on the fly if there's a version mismatch.

Note on compatibility: If you deploy using kubeadm without specifying a version, it will now default to 1.36, which means your host's `kubeadm` package needs to be 1.36+ (or you can explicitly pin an older version in your YAML).


bstoll:
This is still WIP - making draft PR for validating presubmit.

kraney (approved):
lgtm

PR #758, opened 2026-08-25
Shift to using dynamic.Interface for meshnet and lemming

kraney:
This is consistent with what we do for vendor images. (Cisco, Juniper, Nokia, Arista.) It reduces the number of indirect dependencies we pick up on K8s API libraries, which simplifies life within Google's monorepo.

Functionally, it changes nothing.

### DIFF
--- a/topo/node/openconfig/openconfig_test.go
+++ b/topo/node/openconfig/openconfig_test.go
@@ -20,18 +20,20 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


topo/node/openconfig/openconfig_test.go
bstoll, line 183: Unrelated but this row is identical to the one above it - this could be removed.

bstoll (approved):
PR title says lemming & meshnet but this change is only for lemming?

PR #762, opened 2026-08-27
Implement KNE Packet Bridge daemon and CLI subcommand

kraney:
## Summary

Adds the **KNE Packet Bridge**: a daemon that carries raw Layer 2 Ethernet frames
between network segments over a gRPC stream, so that a KNE topology can be wired to
something that is not in the same cluster — another KNE cluster, a workstation `veth`,
or a process running outside Kubernetes entirely.

A `FORWARD` node declares the wires it terminates and KNE derives the daemon
invocations. The topology never names a flag, a port, or a Kubernetes object.

## The contract

`proto/forward.proto` already described this model — a list of wires, each with an `a`
and a `z` endpoint — but nothing implemented it. This PR implements it, including the
`RemoteNode` endpoint that was declared and unreachable.

The endpoint written as an `interface` is always the declaring node:

| Wire | Meaning |
| --- | --- |
| `a: { interface }`, `z: { local_node }` | this node dials a peer in the topology |
| `a: { interface }`, `z: { remote_node }` | this node dials an address outside it |
| `z: { interface }`, no `a` | this node listens; the client is external |

So a node that a Borg job attaches to is written as:

```textproto
nodes: {
  name: "bridge-server"
  vendor: FORWARD
  config: {
    vendor_data: {
      [type.googleapis.com/forward.ForwardConfig] {
        wires: { z: { interface: { name: "eth1" } } }
      }
    }
  }
  services: { key: 50058 value: { name: "wire" inside: 50058 type: NODE_PORT } }
}
```

No image, no CLI flags, no Service names. Two consequences fall out of the model:

- **Peer resolution.** Naming a peer with `local_node` requires that name to resolve,
  and nothing created a Service under a node's own name. Every `FORWARD` node now gets
  a headless Service named after it. It declares no ports — a headless Service needs
  only a selector to publish A records, and ports would place it in the map `kne show`
  reports and collide with the wire port a topology may already expose. It carries the
  `pod` label so the node's existing teardown removes it.
- **Mode per process.** The daemon runs either as a server or as a client, so a node
  that both serves and dials gets one container per role, named `<node>-<role>`. A node
  needing only one keeps the plain node name.

## Changes

**Bridge daemon and client** (`bridge/`, `cmd/bridge/`)
Bidirectional streaming of raw frames over the existing `Wire` service, with a
per-interface demux so one server can serve many wires on one port. `kne bridge server`
and `kne bridge client`, with interface binding, reconnect, and optional ALTS transport
credentials for DirectPath on GCE.

**Wire planning** (`topo/node/forward/`)
`ForwardConfig` is translated into daemon invocations and containers, plus the headless
peer Service described above. The default image points at the published bridge image, so
topologies no longer carry an `image:` line.

**Service exposure** (`proto/topo.proto`, `topo/node/`, `topo/topo.go`)
`Service.Type` gains `LOAD_BALANCER`, `NODE_PORT` and `CLUSTER_IP`, with one Kubernetes
Service created per requested type. `Service.external_ipv6` states that a service should
be reachable over IPv6 from outside the cluster; on a single-stack IPv4 cluster KNE
satisfies that with a host-network `socat` DaemonSet, and on a dual-stack cluster it may
become a no-op. `outside_ip` is now documented as populated only for `LOAD_BALANCER`.

**Build and release** (`Makefile`, `deploy/bridge/`, `cloudbuild/`, `cmd/release/`)
A Dockerfile whose entrypoint is `kne bridge`, `make` targets for local kind testing, and
`kne release bridge` driving a Cloud Build pipeline. The bridge image is the `kne` binary,
so it has no version of its own: releasing it tags KNE as a whole and labels the image
with that version, rather than minting a separate series.

## Testing

Unit tests for `bridge/`, `cmd/bridge/`, the wire planner and peer Service in
`topo/node/forward/`, and the service-type handling in `topo/`. The planner tests cover
each endpoint shape, both malformed-wire cases, and the mixed server-plus-client node.

Verified on kind: `kne deploy`, then `kne create` of `examples/bridge/paired-bridge.pb.txt`,
then `ping` between the two hosts, which have no meshnet link and reach each other only
across the bridge. That exercises wire planning, peer resolution through the headless
Service, stream establishment, and bidirectional forwarding.

Also verified on kind with `external_ipv6: true` set on the service: the `socat` DaemonSet
is admitted, which exercises the `apps`/`daemonsets` grant added to the `podrunner`
ClusterRole, and its pod reaches `Running`, which exercises the explicit non-root UID the
container needs in order for the kubelet to accept it.

Not yet exercised on a cluster: multi-container `FORWARD` nodes and `remote_node`
endpoints. Both are covered by unit tests only.

## Notes for reviewers

- `Service.v6_host_proxy` was renamed to `Service.external_ipv6`. This is a breaking proto
  change, made deliberately while nothing depends on the field, because the old name
  described the mechanism rather than the intent.
- `deploy/ubuntu/serviceaccount.yaml` is reindented throughout. super-linter runs prettier
  over every file a PR touches rather than only the changed lines, and this file did not
  previously indent block sequences under their mapping key. The parsed documents are
  unchanged.
- `x/wire/` contains an earlier, unfinished implementation of the same `Wire` service. It
  is untouched here and nothing imports it; retiring it is left to a follow-up.


bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+		onClose:   onClose,
+		listeners: make(map[chan []byte]struct{}),
+		ctx:       ctx,
+		cancel:    cancel,
+	}
+	// Unblock unix.Recvfrom immediately when context is cancelled.
+	go func() {
+		<-d.ctx.Done()
+		d.closeHandler()
+	}()
```
manizzzz, line 222:
```go
// Unblock unix.Recvfrom immediately when context is cancelled.
go func() {
    <-d.ctx.Done()
    d.closeHandler()
}()
```

This comment is not true on Linux. Closing a file descriptor does **not** wake another thread blocked in `recvfrom(2)` on it — the blocked syscall already holds a reference to the open file description, so it stays blocked, potentially forever. (This is exactly why Go's `net` package uses the runtime netpoller, and why closing a raw fd with a concurrent reader is documented as unsafe.)

Consequences, all reachable in production and none covered by a test:

1. `readLoop` never returns → its goroutine leaks, the `defer` never runs, subscriber channels are never closed, `onClose` is never called, so the demux stays in `s.demuxers` forever.
2. The AF_PACKET socket is never reclaimed — the blocked syscall pins the socket and its `SO_RCVBUF`/`SO_SNDBUF` of 4 MB each (`socketBufferSizeBytes`). ~8 MB of kernel memory per leak.
3. `cmd/bridge/bridge.go:runServer` calls `grpcServer.GracefulStop()` on SIGTERM. `GracefulStop` waits for in-flight `Transmit` handlers; those wait on `errChan`; the egress goroutine waits for `pktChan` to close, which only happens in the `readLoop` defer that never runs. **The daemon does not exit on SIGTERM** and is SIGKILLed after the pod's grace period.

Concrete fix — make the fd interruptible. Either hand it to the runtime netpoller:

```go
func NewSocketHandler(ifaceName string) (*SocketHandler, error) {
    ...
    if err := unix.SetNonblock(fd, true); err != nil {
        _ = unix.Close(fd)
        return nil, fmt.Errorf("failed to set non-blocking on %s: %w", ifaceName, err)
    }
    f := os.NewFile(uintptr(fd), "packet-"+ifaceName) // netpoller now owns fd
    rc, err := f.SyscallConn()
    ...
}

func (s *SocketHandler) ReadPacket() ([]byte, error) {
    for {
        var n int
        var from unix.Sockaddr
        var serr error
        if rerr := s.rc.Read(func(fd uintptr) bool {
            n, from, serr = unix.Recvfrom(int(fd), s.rbuf, 0)
            return serr != unix.EAGAIN // false => park in netpoller, woken by Close()
        }); rerr != nil {
            return nil, rerr // returns ErrClosed once f.Close() runs -> loop exits
        }
        ...
    }
}

func (s *SocketHandler) Close() error {
    var err error
    s.closeOnce.Do(func() { err = s.f.Close() })
    return err
}
```

…or, to keep raw syscalls, the cheap version: set `SO_RCVTIMEO` (e.g. 250 ms) and treat `EAGAIN`/`EINTR` as "loop and re-check `d.ctx`". Either way please fix the misleading comment and document on the `ReadWriter` interface that **`Close` must interrupt a pending `ReadPacket`** — that is the contract `readLoop` and `Client.runStream` silently depend on.

bridge/client.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func (c *Client) runStream(ctx context.Context, conn *grpc.ClientConn) error {
+	handler, err := c.socketOpener(c.cfg.LocalInterface)
+	if err != nil {
+		return fmt.Errorf("failed to open local interface %s: %w", c.cfg.LocalInterface, err)
+	}
+	defer func() {
+		_ = handler.Close()
+	}()
```
manizzzz: Same root cause, but inside an infinite retry loop, so the leak is unbounded:

```go
func (c *Client) runStream(ctx context.Context, conn *grpc.ClientConn) error {
    handler, err := c.socketOpener(c.cfg.LocalInterface)   // a NEW AF_PACKET socket per attempt
    defer func() { _ = handler.Close() }()
    ...
    go func() { ... handler.ReadPacket() ... }()           // blocked here when we return
```

`runStream` returns as soon as one value lands on `errChan` (or `ctx` is done). The egress goroutine is at that moment parked in `handler.ReadPacket()`; per B1 the deferred `Close()` will not wake it. `Run` then sleeps `RetryInterval` (default **2 s**) and calls `runStream` again, opening another socket. A peer that is down or flapping therefore leaks one goroutine + one promiscuous AF_PACKET socket (+8 MB of kernel buffers) every 2 seconds, forever. A bridge client left overnight against a dead peer will OOM the node or hit `EMFILE`.

Fix: B1 makes `Close()` actually interrupt the read. On top of that, make `runStream` wait for its children so the leak cannot come back:

```go
var wg sync.WaitGroup
wg.Add(2)
go func() { defer wg.Done(); ... }()
go func() { defer wg.Done(); ... }()
defer func() { streamCancel(); _ = handler.Close(); wg.Wait() }()
```

Also consider opening the socket **once** in `Run` rather than per reconnect — reopening a promiscuous socket every 2 s is wasteful and racy with interface churn.

bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+			if err != nil {
+				errChan <- err
+				return
+			}
+			if pktData := req.GetData(); len(pktData) > 0 {
+				if err := demux.handler.WritePacket(pktData); err != nil {
+					errChan <- fmt.Errorf("failed to write packet to %s: %w", ifaceName, err)
+					return
+				}
+			}
```
manizzzz, line 499:
```go
go func() {
    for {
        req, err := stream.Recv()
        ...
        if err := demux.handler.WritePacket(pktData); err != nil {
```

`demux.handler` can be closed concurrently by `closeHandler()` (the context-watcher goroutine or `readLoop`'s defer) while this goroutine is inside `unix.Write(s.fd, pkt)`. There is no synchronisation between `SocketHandler.Close` and `SocketHandler.WritePacket`. Best case: `EBADF` and the stream dies. Worst case: the fd number was already recycled by the runtime (another socket, a log file, the gRPC connection itself) between the `Close` and the `Write`, and you inject a raw Ethernet frame into an unrelated file descriptor. Classic use-after-close — and `go test -race` will never catch it, because it is a kernel-level race, not a memory race.

Fix — make the handler self-guarding:

```go
type SocketHandler struct {
    ifaceName string
    mu        sync.RWMutex
    fd        int
    closed    bool
    closeOnce sync.Once
}

func (s *SocketHandler) WritePacket(pkt []byte) error {
    s.mu.RLock()
    defer s.mu.RUnlock()
    if s.closed {
        return net.ErrClosed
    }
    _, err := unix.Write(s.fd, pkt)
    return err
}

func (s *SocketHandler) Close() error {
    var err error
    s.closeOnce.Do(func() {
        s.mu.Lock()
        s.closed = true
        err = unix.Close(s.fd)
        s.fd = -1
        s.mu.Unlock()
    })
    return err
}
```

(The `os.File` variant in B1 gets this for free.) Separately, `Transmit` should not reach into the unexported `demux.handler` field at all — add `func (d *InterfaceDemux) write(pkt []byte) error` that checks `d.ctx.Err()` first.

bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+
+	d := newInterfaceDemux(s.ctx, ifaceName, handler, func(name string) {
+		s.mu.Lock()
+		delete(s.demuxers, name)
+		s.mu.Unlock()
+		klog.Infof("InterfaceDemux for %s cleaned up and removed from server cache", name)
+	})
+	s.demuxers[ifaceName] = d
+	return d, nil
+}
```
manizzzz, line 403: `readLoop`'s teardown order is:

```go
defer func() {
    d.cancel()
    d.closeHandler()
    d.mu.Lock()
    for ch := range d.listeners { close(ch) }
    d.listeners = make(map[chan []byte]struct{})   // (1) listeners reset
    d.mu.Unlock()
    if d.onClose != nil { d.onClose(d.ifaceName) } // (2) removed from s.demuxers
}()
```

Between (1) and (2) — and for the whole duration of a socket read error before (2) — `s.demuxers[iface]` still points at a **dead** demux. A `Transmit` arriving in that window pulls it from the cache, calls `subscribe()`, and gets a channel inserted into the fresh map of a demux whose read loop is gone. That channel is never closed and never written to. The egress goroutine blocks on it; the client sees a successfully established stream (headers were already sent) with **zero packets, forever, and no error**. For a packet bridge, a silent blackhole is worse than a failure.

Verified:

```go
d := newInterfaceDemux(ctx, "eth1", fake, nil)
_ = fake.Close()      // kill readLoop
<-d.ctx.Done()
ch := d.subscribe()   // -> orphan channel
// FAIL: subscribe() on a dead demux returned a channel that is never closed and never written
```

Fix — give the demux an explicit closed state, make `subscribe` fail, and never hand out a dead demux:

```go
func (d *InterfaceDemux) subscribe() (chan []byte, error) {
    d.mu.Lock()
    defer d.mu.Unlock()
    if d.closed {                      // set in readLoop's defer under the same lock
        return nil, fmt.Errorf("demux for %s is closed", d.ifaceName)
    }
    ch := make(chan []byte, channelBufferCap)
    d.listeners[ch] = struct{}{}
    return ch, nil
}

// getOrCreateDemux
if d, ok := s.demuxers[ifaceName]; ok {
    if d.ctx.Err() == nil {
        return d, nil
    }
    delete(s.demuxers, ifaceName)      // dead: fall through and build a fresh one
}
```

and make the teardown callback identity-safe so a late `onClose` can never evict its own replacement:

```go
d := newInterfaceDemux(s.ctx, ifaceName, handler, func(name string) {
    s.mu.Lock()
    if cur, ok := s.demuxers[name]; ok && cur == d {   // pointer identity
        delete(s.demuxers, name)
    }
    s.mu.Unlock()
})
```

`Transmit` then handles the `subscribe` error and returns `status.Errorf(codes.Unavailable, ...)` so the client reconnects instead of hanging.

bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+// Close closes all underlying raw sockets and demuxers.
+func (s *Server) Close() error {
+	s.cancel()
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	for _, d := range s.demuxers {
+		d.close()
+	}
+	return nil
+}
```
manizzzz, line 551:
```go
func (s *Server) Close() error {
    s.cancel()
    ...
    for _, d := range s.demuxers { d.close() }
    return nil          // <- always nil, never waits
}
```

`d.close()` is just `d.cancel()`. `Close()` returns while read loops are still running and sockets are still open, and it discards every `handler.Close()` error (`closeHandler` does `_ = d.handler.Close()`). Verified: immediately after `server.Close()` returns, `len(s.demuxers) == 1`. Any caller that closes a server and then expects the interfaces to be free (tests, `runServer`'s `defer`, a future restart path) is relying on timing.

Fix — track read loops in a `sync.WaitGroup`, wait, and aggregate:

```go
func (s *Server) Close() error {
    s.cancel()
    s.mu.Lock()
    ds := slices.Collect(maps.Values(s.demuxers))
    s.mu.Unlock()
    var errs []error
    for _, d := range ds {
        d.close()
        errs = append(errs, d.wait())   // returns the handler.Close() error
    }
    return errors.Join(errs...)
}
```

bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+	defer s.mu.Unlock()
+	s.socketOpener = opener
+}
+
+// getOrCreateDemux retrieves an existing demuxer for the interface or opens a new one.
+// It attaches a teardown callback so dead demuxers are automatically purged from s.demuxers.
+func (s *Server) getOrCreateDemux(ifaceName string) (*InterfaceDemux, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
```
manizzzz, line 375: There is no `s.ctx.Err()` check, so a `Transmit` racing with shutdown calls `s.socketOpener(...)` and opens a brand-new promiscuous raw socket *after* the server has been closed. The new demux is built with an already-cancelled parent so it self-destructs immediately — but it is also registered in `s.demuxers`, and with the real `SocketHandler` the fd leaks per B1. Verified: `getOrCreateDemux("eth1")` returns a nil error after `server.Close()`.

```go
func (s *Server) getOrCreateDemux(ifaceName string) (*InterfaceDemux, error) {
    s.mu.Lock()
    defer s.mu.Unlock()
    if err := s.ctx.Err(); err != nil {
        return nil, status.Errorf(codes.Unavailable, "bridge server is shutting down: %v", err)
    }
    ...
```

bridge/bridge.go
manizzzz, line 499:
```go
if err := demux.handler.WritePacket(pktData); err != nil {
    errChan <- fmt.Errorf("failed to write packet to %s: %w", ifaceName, err)
    return
}
```

Any single bad frame from the peer kills the entire bridge stream. The most likely case is trivially reachable: a frame larger than the interface MTU returns `EMSGSIZE`, and `maxFrameSize` (65535) is far above a normal 1500-byte MTU, so a peer on a jumbo-frame link will repeatedly knock the stream down and the client will reconnect-loop. Please distinguish fatal (`EBADF`, `ENETDOWN`, `net.ErrClosed`) from per-packet errors; for the latter log + bump a dropped counter (mirroring `droppedFrames` on the egress side) and continue. Also add an explicit length check with a clear error rather than letting the kernel decide.

bridge/bridge.go
```diff
@@ -0,0 +1,410 @@
...
+			d.onClose(d.ifaceName)
+		}
+	}()
+
+	for {
+		select {
+		case <-d.ctx.Done():
+			return
+		default:
+			pkt, err := d.handler.ReadPacket()
```
manizzzz: - `select { case <-d.ctx.Done(): return; default: ReadPacket() }` only observes cancellation *between* frames. On an idle interface, even with B1 fixed, cancellation is noticed only when the next frame arrives. The B1 fix (netpoller / `SO_RCVTIMEO`) resolves this too. Note that with a non-blocking `ReadWriter` implementation this loop is a 100 %-CPU spin.
- `packetPool` is nearly pointless: `ReadPacket` immediately `make`s a right-sized copy anyway, so the pool only recycles the scratch buffer while adding `sync.Pool` overhead per packet. Each `SocketHandler` has exactly one reader — make the 64 KB scratch buffer a field on `SocketHandler` and delete the pool.
- `channelBufferCap = 1000` frames/subscriber × up to 64 KB ≈ 64 MB per subscriber, with no cap on subscribers per interface. Consider bounding by bytes and capping concurrent subscribers.
- The same `pkt` backing array is fanned out to every listener. Safe today (read-only), but please document it on `subscribe()` — an easy future footgun.

topo/node/node.go
```diff
@@ -648,6 +725,12 @@ func (n *Impl) DeleteConfig(ctx context.Context) error {
 
 // DeleteService removes the service definition for the Node.
 func (n *Impl) DeleteService(ctx context.Context) error {
+	_ = n.KubeClient.AppsV1().DaemonSets(n.Namespace).Delete(ctx, fmt.Sprintf("v6proxy-%s", n.Name()), metav1.DeleteOptions{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+		},
+		GracePeriodSeconds: pointer.Int64(0),
+	})
```
manizzzz:
```go
func (n *Impl) DeleteService(ctx context.Context) error {
    _ = n.KubeClient.AppsV1().DaemonSets(n.Namespace).Delete(ctx, fmt.Sprintf("v6proxy-%s", n.Name()), ...)
    return n.KubeClient.CoreV1().Services(n.Namespace).Delete(...)
}
```

The `_ =` discards everything: RBAC `Forbidden`, API timeouts, `context.Canceled`, conflicts — all indistinguishable from the expected `NotFound`. When it fails, a `hostNetwork` DaemonSet keeps running socat on every node in the cluster, bound to the host's IPv6 NodePort. Nothing is logged, `kne topology delete` reports success, and the next topology that lands on the same NodePort fails with an opaque `address already in use` from a socat container nobody knows exists. Manual `kubectl delete ds -n <ns>` is the only recovery.

Verified: with a reactor returning `Forbidden` on `delete daemonsets`, `DeleteService` returns `nil`.

Fix — ignore only `NotFound`, and still attempt the Service delete:

```go
func (n *Impl) DeleteService(ctx context.Context) error {
    var errs []error
    dsName := fmt.Sprintf("v6proxy-%s", n.Name())
    if err := n.KubeClient.AppsV1().DaemonSets(n.Namespace).Delete(ctx, dsName, metav1.DeleteOptions{
        GracePeriodSeconds: pointer.Int64(0),
    }); err != nil && !apierrors.IsNotFound(err) {
        errs = append(errs, fmt.Errorf("failed to delete v6proxy daemonset %q: %w", dsName, err))
    }
    if err := n.KubeClient.CoreV1().Services(n.Namespace).Delete(ctx, fmt.Sprintf("service-%s", n.Name()), metav1.DeleteOptions{
        GracePeriodSeconds: pointer.Int64(0),
    }); err != nil {
        errs = append(errs, err)
    }
    return errors.Join(errs...)
}
```

Related (N11): `Impl.Delete` downgrades this to `log.Warningf` and returns `nil` anyway, so even the fixed error stays invisible to `kne topology delete`.

topo/node/node.go
```diff
@@ -571,6 +573,14 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
 		}
 		servicePorts = append(servicePorts, sp)
+		switch v.Type {
+		case tpb.Service_NODE_PORT:
+			svcType = corev1.ServiceTypeNodePort
+		case tpb.Service_CLUSTER_IP:
+			if svcType != corev1.ServiceTypeNodePort {
+				svcType = corev1.ServiceTypeClusterIP
+			}
+		}
```
manizzzz:
```go
svcType := corev1.ServiceTypeLoadBalancer
for k, v := range n.Proto.Services {
    ...
    switch v.Type {
    case tpb.Service_NODE_PORT:
        svcType = corev1.ServiceTypeNodePort
    case tpb.Service_CLUSTER_IP:
        if svcType != corev1.ServiceTypeNodePort {
            svcType = corev1.ServiceTypeClusterIP
        }
    }
}
```

A Kubernetes `Service` has exactly **one** `spec.type` covering all of its ports, but `Service.type` in the proto is **per port**. This loop collapses N per-port intents into one value, so one port silently rewrites the exposure of every other port on the node.

Verified failures on this tree:

1. Node with `ssh:22` (unset → LoadBalancer), `gnmi:9339` (`LOAD_BALANCER`) and `wire:50058` (`NODE_PORT`) → the created Service is `type: NodePort` with all three ports.
   - Every port now gets a NodePort allocated, including SSH and gNMI. The original code deliberately set `AllocateLoadBalancerNodePorts: false` with a comment warning that *"Large topologies may try to allocate more NodePorts than are supported in default clusters"* — and the new `if svcType == LoadBalancer` guard skips that protection entirely for this Service. The default node-port range is ~2768 ports; a large topology will exhaust it.
   - The node also loses its LoadBalancer ingress IP, so per T1 `outside_ip` becomes `""` for **all** of its services — a silent breaking change for anything consuming `kne show` / the topology proto (ondatra bindings, test harnesses).
2. Node with `gnmi:9339` (`LOAD_BALANCER`) + `http:8080` (`CLUSTER_IP`) → the created Service is `type: ClusterIP`. An explicitly requested `LOAD_BALANCER` port is silently downgraded and becomes unreachable from outside. Note `tpb.Service_LOAD_BALANCER` has **no case in the switch at all**, so an explicit LB request cannot defend itself.
3. Not reproducible with `kfake` (it performs no validation) but fatal on a real cluster: if any port sets `node_port` and the collapsed type ends up `ClusterIP`, the API server rejects the whole create with `may not be used when type is 'ClusterIP'`, and the failure is attributed to the wrong port.

Preferred fix — stop collapsing; group ports by type and create one Service per group:

```go
groups := map[corev1.ServiceType][]corev1.ServicePort{}
for k, v := range n.Proto.Services {
    t := k8sType(v.GetType())            // UNKNOWN -> LoadBalancer
    groups[t] = append(groups[t], sp)
}
// service-<node>            -> LoadBalancer (unchanged name, preserves back-compat)
// service-<node>-nodeport   -> NodePort
// service-<node>-clusterip  -> ClusterIP
```

(`populateServiceMap` already keys by port, so it merges multiple Services per node correctly — just make sure the node's Service lookup/list selector picks up all of them.)

Minimum acceptable fix if you don't want multiple Services in this PR: **detect the conflict and fail loudly** instead of silently picking a winner, and handle `LOAD_BALANCER` explicitly:

```go
var want corev1.ServiceType
for k, v := range n.Proto.Services {
    t := k8sType(v.GetType())
    if want != "" && want != t {
        return fmt.Errorf("node %q: service port %d requests type %q but another port on the same node requests %q; "+
            "a Kubernetes Service has a single type, so all services on a node must agree", n.Name(), k, t, want)
    }
    want = t
}
```

Please also add the mixed-type cases to `TestService` — today every new case has exactly one service, which is precisely why this passes review.

topo/node/node.go
```diff
@@ -588,20 +598,87 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+		if svc, ok := n.Proto.Services[uint32(sp.Port)]; ok && svc.GetV6HostProxy() {
+			needsProxy = true
+		} else {
+			for _, svc := range n.Proto.Services {
+				if svc.GetName() == sp.Name && svc.GetV6HostProxy() {
+					needsProxy = true
+					break
+				}
+			}
+		}
```
manizzzz:
```go
if svc, ok := n.Proto.Services[uint32(sp.Port)]; ok && svc.GetV6HostProxy() {
    needsProxy = true
} else {
    for _, svc := range n.Proto.Services {
        if svc.GetName() == sp.Name && svc.GetV6HostProxy() {
            needsProxy = true
            break
        }
    }
}
```

`sp.Port` is `int32(k)` where `k` is the map key, so the keyed lookup always succeeds — the fallback is dead weight in the happy path and actively wrong otherwise. `name` is optional and not unique. When the keyed service exists but has `v6_host_proxy: false`, control falls into the `else`, and if any other service shares its name (including the very common case of **two unnamed services**, `"" == ""`) it gets a proxy it never asked for.

A node with `50058 {v6_host_proxy: true}` and `50059 {}`, both unnamed, produces a DaemonSet with **two** socat containers (`socat-30058`, `socat-30059`) — an extra host IPv6 port opened for a service that explicitly did not request one.

Fix — delete the fallback entirely:

```go
if svc, ok := n.Proto.Services[uint32(sp.Port)]; ok && svc.GetV6HostProxy() && sp.NodePort > 0 {
    proxyContainers = append(proxyContainers, ...)
}
```

While you're there: `needsProxy && sp.NodePort == 0` is currently a silent no-op. If a user sets `v6_host_proxy: true` on a `CLUSTER_IP` service (or the type collapsed per N2), they get no proxy and no message. Please `log.Warningf` (or error) in that branch.

topo/node/node.go
```diff
@@ -588,20 +598,87 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+						TerminationGracePeriodSeconds: pointer.Int64(0),
+					},
+				},
+			},
+		}
+		sProxyDS, err := n.KubeClient.AppsV1().DaemonSets(n.Namespace).Create(ctx, proxyDS, metav1.CreateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to create host v6proxy daemonset: %w", err)
+		}
+		log.Infof("Created host v6proxy daemonset:\n%+v\n", sProxyDS)
```
manizzzz, line 774: If `DaemonSets().Create()` fails, the function returns an error but the `Service` it just created is left behind. The node is half-created, and with no `AlreadyExists` handling a retry of `CreateService` fails on the Service create with a confusing error instead of converging. Either delete the Service on the DaemonSet failure path (named-return `defer` rollback), or make both creates tolerate `AlreadyExists` and update in place.

topo/node/node.go
```diff
@@ -588,20 +598,87 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+				},
+				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: dsLabels,
+					},
+					Spec: corev1.PodSpec{
+						HostNetwork:                   true,
+						Containers:                    proxyContainers,
+						TerminationGracePeriodSeconds: pointer.Int64(0),
+					},
```
manizzzz, line 766: - **No `OwnerReference`.** The DaemonSet is only ever removed by the exact-name delete in `DeleteService`. Set `OwnerReferences` to the Service (or the node's Pod) so the garbage collector cleans it up when the owner disappears — belt to N1's braces.
- **No `Tolerations`/`NodeSelector`.** A DaemonSet will not schedule onto tainted nodes (e.g. `node-role.kubernetes.io/control-plane:NoSchedule`). On a kind cluster that is usually exactly the node you hit the NodePort on, so the IPv6 path silently won't work there. Add the standard control-plane tolerations, or `nodeSelector` the proxy onto the nodes that matter.
- **No `DNSPolicy: ClusterFirstWithHostNet`** on a `hostNetwork: true` pod. Not needed for a literal-IP socat today, but it is the standard idiom and avoids surprises if the args ever become DNS names.
- `TerminationGracePeriodSeconds: 0` means SIGKILL with no grace; combined with `fork` that can orphan children mid-connection. Worth a comment saying it's intentional.

topo/node/node.go
```diff
@@ -588,20 +598,87 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+				if svc.GetName() == sp.Name && svc.GetV6HostProxy() {
+					needsProxy = true
+					break
+				}
+			}
+		}
+		if needsProxy && sp.NodePort > 0 {
+			proxyContainers = append(proxyContainers, corev1.Container{
+				Name:  fmt.Sprintf("socat-%d", sp.NodePort),
+				Image: "alpine/socat:latest",
```
manizzzz:
```go
Image: "alpine/socat:latest",
ImagePullPolicy: corev1.PullIfNotPresent,
```

Three problems in two lines: (1) a hardcoded image string buried in `CreateService` — every other image in KNE is configurable; (2) a floating `:latest` tag from Docker Hub — unreproducible across nodes, subject to anonymous pull rate limits, and a supply-chain risk for something running with `hostNetwork`; (3) `PullIfNotPresent` + `latest` means different nodes can durably run different builds. Air-gapped/offline clusters (a supported KNE use case) cannot pull it at all.

topo/node/node.go
```diff
@@ -588,20 +598,87 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+				Args: []string{
+					fmt.Sprintf("TCP6-LISTEN:%d,fork,reuseaddr", sp.NodePort),
```
manizzzz: `TCP6-LISTEN:<port>,fork,reuseaddr` with no `max-children` lets any reachable client fork unbounded socat processes (with a 64Mi limit the cgroup starts OOM-killing under a trivial connection flood). Add `max-children=<N>`, and add a `SecurityContext` — the ports are >1024 so no capabilities are required:

```go
SecurityContext: &corev1.SecurityContext{
    AllowPrivilegeEscalation: pointer.Bool(false),
    Capabilities:             &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}},
    RunAsNonRoot:             pointer.Bool(true),
},
```

Also set `Requests` explicitly instead of relying on defaulting from `Limits`, so the QoS class is intentional.

topo/topo.go
```diff
@@ -942,7 +942,7 @@ var populateServiceMap = func(s *corev1.Service, m map[uint32]*tpb.Service) erro
 	if s == nil || m == nil {
 		return fmt.Errorf("service and map must not be nil")
 	}
-	if len(s.Status.LoadBalancer.Ingress) == 0 {
+	if (s.Spec.Type == corev1.ServiceTypeLoadBalancer || s.Spec.Type == "") && len(s.Status.LoadBalancer.Ingress) == 0 {
```
manizzzz:
```go
if (s.Spec.Type == corev1.ServiceTypeLoadBalancer || s.Spec.Type == "") && len(s.Status.LoadBalancer.Ingress) == 0 {
```

1. An empty `spec.type` means **ClusterIP** in Kubernetes, not LoadBalancer — the API server always defaults it, so `""` only ever appears on fakes/hand-built objects. Treating `""` as LoadBalancer preserves the old fixtures but encodes the wrong semantic. Either normalise once (`t := s.Spec.Type; if t == "" { t = corev1.ServiceTypeClusterIP }`) and fix the fixtures, or keep the behaviour with a comment saying it exists only for legacy/fake objects.
2. `service.OutsideIp` is now left `""` for NodePort/ClusterIP services. `outside_ip` is part of the public topology proto that `kne show` prints and that downstream harnesses dial; an empty IP will surface as `dial tcp :50058` far from here. For NodePort, consider populating the node IP (`Node.Status.Addresses[ExternalIP/InternalIP]`) — the actually reachable address — or explicitly document/validate the empty value. Either way this is an observable output change that deserves a release note.

bridge/bridge_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (f *fakeReadWriter) Close() error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if !f.closed {
+		f.closed = true
+		close(f.readChan)
+	}
+	return nil
+}
```
manizzzz, line 79: nit:  The fake's `Close()` unblocks `ReadPacket`; the real one can't, so the teardown tests assert a property production lacks.

bridge/bridge_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (f *fakeReadWriter) WritePacket(pkt []byte) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if f.writeErr != nil {
+		return f.writeErr
+	}
+	f.writeChan <- pkt
+	return nil
+}
```
manizzzz, line 69: Declared and never assigned → the ingress write-error path is untested

bridge/bridge_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (f *fakeReadWriter) isClosed() bool {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.closed
+}
+
+func TestTransmitBidirectionalStream(t *testing.T) {
```
manizzzz, line 97: Only ever one client; the whole point of the demux (fan-out to N clients) is untested.
The entire justification for `InterfaceDemux` is fanning one socket out to N clients, yet there is no test with two concurrent `Transmit` streams on the same interface asserting both receive the same frame. `TestDemuxerSlowSubscriberDoesNotDeadlock` pokes the channels directly, which is not the same thing.

bridge/bridge_test.go
```diff
@@ -0,0 +1,340 @@
...
+
+	server := NewServer(ctx)
+	defer func() {
+		_ = server.Close()
+	}()
+
+	fakeIO := newFakeReadWriter()
+	server.SetSocketOpener(func(ifaceName string) (ReadWriter, error) {
+		return fakeIO, nil
+	})
```
manizzzz, line 109: Assert `socketOpener` is invoked exactly once for two sequential/concurrent `Transmit` calls on the same interface (and once per distinct interface). Note the current tests return the *same* `fakeIO` regardless of `ifaceName`, so a two-interface test would double-`Close` one fake — the fake should be per-interface.

bridge/bridge_test.go
```diff
@@ -0,0 +1,340 @@
...
+	case receivedPkt := <-fakeIO.writeChan:
+		if !bytes.Equal(receivedPkt, ingressPacket) {
+			t.Fatalf("Ingress packet mismatch: got %v, want %v", receivedPkt, ingressPacket)
+		}
+	case <-time.After(1 * time.Second):
+		t.Fatalf("Timed out waiting for ingress packet to be written to raw socket")
+	}
+}
+
+func TestDemuxerSlowSubscriberDoesNotDeadlock(t *testing.T) {
```
manizzzz, line 171: **No test for `onClose` cache eviction or for reconnect after a read error.**. Regression test (fails today):

```go
d := newInterfaceDemux(ctx, "eth1", fake, nil)
_ = fake.Close()
<-d.ctx.Done()
time.Sleep(100 * time.Millisecond)
if _, err := d.subscribe(); err == nil {
    t.Error("subscribe on a dead demux must fail")
}
```

manizzzz, line 171: Nothing runs `subscribe`/`unsubscribe`/`close`/`readLoop` concurrently, so the `-race` run is basically decorative. Add a fan-in/fan-out stress test (N goroutines subscribing/unsubscribing while frames flow) and run with `-race -count=10`.

manizzzz (commented):
Thanks for the changes.

topo/topo.go
kraney: Revised to match K8s semantics for clusterIP. Outside_ip is populated by metallb in cases where it's relevant. NodePort services don't get an outside IP by definition; the IP is a cluster node's IP in that case.

bridge/bridge_test.go
kraney: Added bridge_test.go:723-760 testing concurrent
  subscribe/unsubscribe and packet flow under -race -count=10
  
  Added node_test.go:607-675 and node_test.go:677-715 verifying DaemonSet attributes and atomic
  rollback on creation failure.

bridge/bridge.go
```diff
@@ -0,0 +1,551 @@
...
+		return fmt.Errorf("packet size %d exceeds max frame size %d", len(pkt), maxFrameSize)
+	}
+	var serr error
+	if werr := s.rc.Write(func(fd uintptr) bool {
+		_, serr = unix.Write(int(fd), pkt)
+		return serr != unix.EAGAIN
+	}); werr != nil {
+		return werr
+	}
+	return serr
```
manizzzz, line 171:
```go
		if rerr := s.rc.Read(func(fd uintptr) bool {
			n, from, serr = unix.Recvfrom(int(fd), s.rbuf, 0)
			return serr != unix.EAGAIN
		}); rerr != nil {
			return nil, rerr
		}
		if serr != nil {
			return nil, serr
		}
```

Go's `internal/poll.FD.RawRead` and `RawWrite` invoke the callback directly and do **not** wrap it in `ignoringEINTR`. On Linux, Go's signal-based async goroutine preemption (`SIGURG`) can interrupt `unix.Recvfrom` / `unix.Write` with `unix.EINTR`.

When that happens:
1. `serr` is `unix.EINTR`.
2. `serr != unix.EAGAIN` evaluates to `true`, so `s.rc.Read` considers the syscall complete and returns `rerr = nil`.
3. `if serr != nil { return nil, serr }` returns `EINTR` to `readLoop()`, which treats any non-nil error as a fatal socket failure (`bridge/bridge.go` L269), **tearing down the entire `InterfaceDemux` and closing all active subscriber streams**.

**Fix:** Retry `unix.EINTR` inside the `s.rc.Read` and `s.rc.Write` callbacks:

```go
		if rerr := s.rc.Read(func(fd uintptr) bool {
			for {
				n, from, serr = unix.Recvfrom(int(fd), s.rbuf, 0)
				if serr == unix.EINTR {
					continue
				}
				return serr != unix.EAGAIN
			}
		}); rerr != nil {
			return nil, rerr
		}
```

And in `WritePacket` (`L165–L171`):

```go
	if werr := s.rc.Write(func(fd uintptr) bool {
		for {
			_, serr = unix.Write(int(fd), pkt)
			if serr == unix.EINTR {
				continue
			}
			return serr != unix.EAGAIN
		}
	}); werr != nil {
		return werr
	}
```

bridge/bridge.go
```diff
@@ -0,0 +1,551 @@
...
+		}
+	}()
+
+	// Wait for stream termination or error.
+	err = <-errChan
+	if err == io.EOF || stream.Context().Err() != nil {
+		return nil
+	}
+	return err
+}
```
manizzzz, line 509: In `bridge/client.go` (`runStream`), a `sync.WaitGroup` was added so `runStream` waits for both the egress and ingress goroutines before returning. However, `Server.Transmit` still returns immediately on the first error from `errChan`:

```go
	// Wait for stream termination or error.
	err = <-errChan
	if err == io.EOF || stream.Context().Err() != nil {
		return nil
	}
	return err
```

If the egress goroutine exits first (e.g. `pktChan` is closed because the demux shut down, or `stream.Send` fails) while the ingress goroutine is still running (or vice versa when `demux.write` hits a fatal error), `Transmit` returns from the RPC handler while the sibling goroutine is still calling `stream.Recv()` or `stream.Send()`. That violates `grpc.ServerStream`'s contract (*"It is not safe to call `SendMsg` or `RecvMsg` after the handler returns"*) and races with gRPC-Go tearing down the server stream.

**Fix:** Unsubscribe `pktChan` (which unblocks the egress goroutine if the ingress goroutine failed first) and wait for the egress goroutine to finish before returning:

```go
	var egressWG sync.WaitGroup
	egressWG.Add(1)
	go func() {
		defer egressWG.Done()
		...
	}()
	defer func() {
		demux.unsubscribe(pktChan) // closes pktChan so egress goroutine unblocks and exits
		egressWG.Wait()
		klog.Infof("Wire.Transmit client disconnected from interface %q", ifaceName)
	}()
```

topo/node/forward/forward.go
```diff
@@ -124,22 +228,50 @@ func (n *Node) CreatePod(ctx context.Context) error {
...
-			}
-			fwdArgs = append(fwdArgs, arg)
+		planned, err := wirePlan(fwdCfg)
+		if err != nil {
+			return fmt.Errorf("node %s: %w", pb.Name, err)
+		}
+		if len(planned) > 0 {
+			procs = planned
+		}
+	}
```
manizzzz, line 247:
```go
	procs := []bridgeProcess{{args: pb.Config.Args}}
	if vendorData := pb.Config.GetVendorData(); vendorData != nil {
		...
		if len(planned) > 0 {
			procs = planned
		}
	}
```

When `ForwardConfig.wires` is used, `pb.Config.Args` is completely discarded and replaced by `planned`. Because `wirePlan` only emits `--peer`, `--interface`, and `--remote_interface` (for client) or `["server"]` (for server), **there is no way to enable `--alts` (or configure `--retry_interval` / `--listen_port`) on a `FORWARD` node configured via `wires`**. Even if a user specifies `config: { args: ["--alts"] }`, those extra flags are silently dropped and the bridge runs insecurely.

**Fix:** Either append `pb.Config.Args` (extra flags like `--alts`) to each generated `bridgeProcess.args` in `CreatePod`:

```go
		if len(planned) > 0 {
			for i := range planned {
				planned[i].args = append(planned[i].args, pb.Config.Args...)
			}
			procs = planned
		}
```

and/or add `bool alts` to `ForwardConfig` in `proto/forward.proto`.

topo/node/forward/forward.go
```diff
@@ -124,22 +228,50 @@ func (n *Node) CreatePod(ctx context.Context) error {
...
+		if len(procs) > 1 {
+			name = fmt.Sprintf("%s-%s", pb.Name, p.nameSuffix)
 		}
+		log.Infof("Container %q args: %v", name, p.args)
+		containers = append(containers, corev1.Container{
+			Name:            name,
+			Image:           pb.Config.Image,
+			Command:         pb.Config.Command,
+			Args:            p.args,
+			Env:             node.ToEnvVar(pb.Config.Env),
```
manizzzz, line 267:
```go
		name := pb.Name
		if len(procs) > 1 {
			name = fmt.Sprintf("%s-%s", pb.Name, p.nameSuffix)
		}
```

When `len(procs) > 1` (e.g. a node that both serves an interface and dials a peer, or dials two interfaces), **none** of the containers in the Pod is named `pb.Name` (they are named `<node>-server`, `<node>-eth1`, etc.). However:
1. `n.Impl.Exec` (`topo/node/node.go` L881) hardcodes `Container: n.Name()`, so any `Exec` call on a multi-wire `FORWARD` node will fail with `container <node> is not valid for pod <node>`.
2. `defaults()` (`forward.go` L351) sets `EntryCommand` to `kubectl exec -it <node> -- sh`.

**Fix:** Keep `pb.Name` for the first container (`i == 0`) and only append `-<nameSuffix>` for `i > 0`:

```go
	for i, p := range procs {
		name := pb.Name
		if i > 0 {
			name = fmt.Sprintf("%s-%s", pb.Name, p.nameSuffix)
		}
```

---

topo/node/forward/forward.go
```diff
@@ -77,41 +84,138 @@ func (n *Node) Create(ctx context.Context) error {
...
+// Within a Wire, the Interface endpoint is always this node: if it is the "a"
+// (client) endpoint then this node dials out, and if it is the "z" (server)
+// endpoint then this node listens and the peer dials in. A wire with no "a" at
+// all is a server wire whose client lives outside the topology entirely, which
+// is how an external peer such as a Borg job attaches.
+//
+// One bridge server handles every interface asked of it over a single port, so
+// all server wires collapse into one process, whereas each client wire dials a
+// distinct peer and needs its own.
+func wirePlan(cfg *fpb.ForwardConfig) ([]bridgeProcess, error) {
```
manizzzz, line 153: `wirePlan` checks that `a` and `z` are not both interfaces on the *same* `Wire`, but it does not check for collisions **across** wires in `cfg.GetWires()`:
1. If two client wires specify the same `a.interface.name` (e.g. `eth1`), `procs` gets two entries with `nameSuffix: "eth1"`, producing duplicate container names (`<node>-eth1`) in the Pod spec (rejected by `kube-apiserver`) and two bridge clients opening raw promiscuous sockets on the same interface.
2. If one wire uses `eth1` as `a.interface` (client) and another wire on the same node uses `eth1` as `z.interface` (server), both the client and server containers open promiscuous sockets on `eth1`, creating an immediate L2 packet loop.
3. If an interface is named `"server"`, its `nameSuffix` collides with the server container's `nameSuffix: "server"`.

**Fix:** Track seen local interfaces in a `map[string]struct{}` inside `wirePlan` and return an error on duplicates; also prefix client container suffixes with `client-` (e.g. `nameSuffix: "client-" + aIntf`) so they can never collide with `"server"`.

topo/node/forward/forward.go
```diff
@@ -77,41 +84,138 @@ func (n *Node) Create(ctx context.Context) error {
...
-	return fmt.Sprintf("--interfaces=%s", intf)
+// CreateService creates the services declared in the topology, plus a headless
+// Service named after the node so that peers naming it with local_node can
+// reach it as "<node>:<wirePort>".
+func (n *Node) CreateService(ctx context.Context) error {
+	if err := n.createPeerService(ctx); err != nil {
+		return err
+	}
+	return n.Impl.CreateService(ctx)
 }
```
manizzzz, line 95:
```go
func (n *Node) CreateService(ctx context.Context) error {
	if err := n.createPeerService(ctx); err != nil {
		return err
	}
	return n.Impl.CreateService(ctx)
}
```

If `n.createPeerService(ctx)` succeeds and `n.Impl.CreateService(ctx)` subsequently fails, `Impl.CreateService` rolls back its own services (`service-<node>*`), but the headless Service `<node>` created by `createPeerService` is left behind in the namespace, causing retries to fail with `AlreadyExists`.

**Fix:** Roll back the headless service if `n.Impl.CreateService(ctx)` returns an error:

```go
func (n *Node) CreateService(ctx context.Context) error {
	if err := n.createPeerService(ctx); err != nil {
		return err
	}
	if err := n.Impl.CreateService(ctx); err != nil {
		_ = n.KubeClient.CoreV1().Services(n.Namespace).Delete(ctx, n.Name(), metav1.DeleteOptions{})
		return err
	}
	return nil
}
```

topo/node/node.go
```diff
@@ -570,58 +620,183 @@ func (n *Impl) CreateService(ctx context.Context) error {
...
+			}
+		}
+	}
+	if len(proxyContainers) > 0 {
+		dsLabels := map[string]string{
+			"app":  fmt.Sprintf("v6proxy-%s", n.Name()),
+			"pod":  n.Name(),
+			"topo": n.Namespace,
+		}
+		var ownerRefs []metav1.OwnerReference
```
manizzzz, line 728:
```go
		var ownerRefs []metav1.OwnerReference
		if len(createdServices) > 0 {
			ownerRefs = []metav1.OwnerReference{
				{
					APIVersion:         "v1",
					Kind:               "Service",
					Name:               createdServices[0].Name,
					UID:                createdServices[0].UID,
...
```

Because `typeOrder` creates `ServiceTypeLoadBalancer` first (`L636–L640`), on a node with mixed services (e.g., `ssh:22` as `LOAD_BALANCER` and `wire:50058` as `NODE_PORT` with `external_ipv6: true`), `createdServices[0]` is `service-<node>` (the `LoadBalancer` Service), **not** `service-<node>-nodeport` (the `NodePort` Service whose ports `v6proxy-<node>` actually forwards).

**Fix:** Record the `sS` from which `proxyContainers` were generated (or prefer the `ServiceTypeNodePort` service in `createdServices`) and use that Service as the controller `OwnerReference`:

```go
		ownerSvc := createdServices[0]
		for _, sS := range createdServices {
			if sS.Spec.Type == corev1.ServiceTypeNodePort {
				ownerSvc = sS
				break
			}
		}
```

topo/node/forward/forward.go
```diff
@@ -77,41 +84,138 @@ func (n *Node) Create(ctx context.Context) error {
...
+	case e.GetRemoteNode() != nil:
+		rn := e.GetRemoteNode()
+		addr := rn.GetAddr()
+		if addr == "" {
+			return "", "", fmt.Errorf("remote_node endpoint must set addr")
 		}
+		if _, _, err := net.SplitHostPort(addr); err != nil {
+			// No port given, so assume the peer serves on the default port.
+			addr = net.JoinHostPort(addr, strconv.Itoa(wirePort))
+		}
```
manizzzz, line 215:
```go
		if _, _, err := net.SplitHostPort(addr); err != nil {
			// No port given, so assume the peer serves on the default port.
			addr = net.JoinHostPort(addr, strconv.Itoa(wirePort))
		}
```

If a user passes a bracketed IPv6 literal without a port (e.g. `addr: "[2001:db8::1]"`), `net.SplitHostPort` fails with `missing port in address`, and `net.JoinHostPort("[2001:db8::1]", "50058")` produces `"[[2001:db8::1]]:50058"` (double brackets), which `grpc.NewClient` cannot dial.

**Fix:** Strip surrounding brackets before calling `net.JoinHostPort`:

```go
		if _, _, err := net.SplitHostPort(addr); err != nil {
			addr = net.JoinHostPort(strings.Trim(addr, "[]"), strconv.Itoa(wirePort))
		}
```

manizzzz (commented):
Thanks for the super thorough fixes on the and new   test suite all look great!
  
  Holding LGTM for three small issues in the new diff before we merge:
  
  1. bridge/bridge.go (ReadPacket / WritePacket): retry unix.EINTR inside the s.rc.Read / s.rc.Write callbacks (RawConn doesn't ignore EINTR, so
  Go's SIGURG preemption will cause readLoop to tear down the demux).
  2. bridge/bridge.go (Server.Transmit): wait for the egress goroutine after demux.unsubscribe(pktChan) in the defer so stream.Send cannot race 
  with Transmit returning.
  3. topo/node/forward/forward.go (CreatePod): preserve pb.Config.Args when ForwardConfig.wires is used so flags like --alts aren't dropped.    
  
  Left a few optional/non-blocking comments on forward.go and node.go as well. Once (1)–(3) are in, this is good to LGTM!

bridge/bridge.go
kraney: Done

_Co-authored by AI Agent_

bridge/bridge.go
kraney: Done

_Co-authored by AI Agent_

topo/node/forward/forward.go
kraney: Done

_Co-authored by AI Agent_

topo/node/forward/forward.go
kraney: Done

_Co-authored by AI Agent_

topo/node/forward/forward.go
kraney: Done

_Co-authored by AI Agent_

topo/node/forward/forward.go
kraney: Done

_Co-authored by AI Agent_

topo/node/node.go
kraney: Done

_Co-authored by AI Agent_

topo/node/forward/forward.go
kraney: Done

_Co-authored by AI Agent_

bstoll (approved):
I didn't verify this but if GSO/TSO are enabled on the veth do TCP and UDP packets get forwarded with valid checksums? I think the way this reads packets they will forward with invalid checksums unless hardware offloads are disabled.

PR #763, opened 2026-08-27
    This PR improves error handling, certificate idempotency, and command response validation for JunOS / cPTX nodes in KNE. 

manizzzz:
1. **Premature Success Validation**: When pushing self-signed certificates via CLI, `waitCertInfraReadyAndPushCert` sent multiple commands  
  (`generate-key-pair` and `generate-self-signed`). The loop previously evaluated `strings.Contains(resp.Result, "successfully")` on any        
  response in the batch. Because key pair generation finishes first with `"Generated key pair ... successfully"`, the method would return `nil` 
  prematurely even if the second command (`generate-self-signed`) failed or was not yet processed.
    2. **Failure on Node Restarts / Reuse**: When a pod/node restarted or was reused with an existing certificate, JunOS returned `error:       
  Command aborted as key pair/certificate already exists`. The previous logic treated this as an error, causing a retry loop until the 15-minute
  timeout failed the deployment.
    3. **Regex Accuracy**: Error checking relied on a simple substring match on `"error:"`. This was vulnerable to false negatives on multi-line
  error blocks or non-standard error phrasings (`commit error`, `syntax error`), as well as potential false positives on configuration strings  
  (e.g., interface descriptions or domain names).
    4. **Log Mismatch**: In `waitConfigInfraReadyAndPushConfigs`, the retry log incorrectly reported `certGenRetrySleep` instead of             
  `configModeRetrySleep`.
  
    ## Changes Made
    - **Idempotent Cert Handling**: Updated `waitCertInfraReadyAndPushCert` to recognize `"already exists"` outputs as a successful / idempotent
  state rather than an error condition.
    - **Dynamic Response Verification**: Dynamically locate the response corresponding specifically to the `generate-self-signed` command,      
  verifying that no non-idempotent errors occurred across any command (`!hasError`) and that certificate generation explicitly succeeded        
  (`certGenSuccess`).
    - **Incomplete Batch Guard**: Added a response count check (`len(multiresp.Responses) < len(commands)`) to avoid evaluating truncated       
  command batches.
    - **Robust Error Regex**: Implemented `isJunosError` with regex `(?im)(\berror:|\bcommit error\b|\bsyntax error\b)` to handle multi-line and
  case-insensitive JunOS CLI errors without false positives.
    - **Log Correction**: Fixed the sleep duration reported in the retry log inside `waitConfigInfraReadyAndPushConfigs`.
    - **Unit Tests & Fixtures**:
      - Added `TestIsJunosError` covering edge cases and false-positive prevention.
      - Updated `TestGenerateSelfSigned` to verify idempotency on existing certificates.
      - Added `testdata/generate_certificate_key_success_cert_failure` test fixture to prevent regressions where key-generation success masks   
  cert-generation failure.
  
    ## Testing Done
    - Ran package unit tests: `go test -v ./topo/node/juniper/...` (100% PASS)
    - Ran all node vendor tests: `go test ./topo/node/...` (100% PASS)
    - Verified formatting with `gofmt` and static checks with `go vet`.

### DIFF
--- a/topo/node/juniper/juniper.go
+++ b/topo/node/juniper/juniper.go
@@ -225,6 +225,12 @@ func (n *Node) GRPCConfig() []string {
 	}
 }
 
+var junosErrorRegex = regexp.MustCompile(`(?im)(\berror:|\bcommit error\b|\bsyntax error\b)`)
+
+func isJunosError(result string) bool {
+	return junosErrorRegex.MatchString(result)
+}
+
 // Waits and retries until CLI config mode is up and config is applied
 func (n *Node) waitConfigInfraReadyAndPushConfigs(configs []string) error {
 	log.Infof("Waiting for config to be pushed (timeout: %v) node %s", configModeTimeout, n.Name())
@@ -246,8 +252,8 @@ func (n *Node) waitConfigInfraReadyAndPushConfigs(configs []string) error {
 					log.Infof("Config mode ready. Config commit done. Node %s", n.Name())
 					return nil
 				}
-				if strings.Contains(resp.Result, "error:") {
-					log.Infof("Config mode not ready. Retrying in %v. Node %s Response %s", certGenRetrySleep, n.Name(), multiresp.JoinedResult())
+				if isJunosError(resp.Result) {
+					log.Infof("Config mode not ready. Retrying in %v. Node %s Response %s", configModeRetrySleep, n.Name(), multiresp.JoinedResult())
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


topo/node/juniper/juniper.go
```diff
@@ -281,18 +287,38 @@ func (n *Node) waitCertInfraReadyAndPushCert() error {
...
 		for _, resp := range multiresp.Responses {
 			if resp.Failed != nil {
 				return resp.Failed
 			}
-			if strings.Contains(resp.Result, "successfully") {
-				log.Infof("Cert Infra ready. Configured Certs. Node %s, Response %s", n.Name(), multiresp.JoinedResult())
-				return nil
-			}
-			if strings.Contains(resp.Result, "error:") {
+			if isJunosError(resp.Result) && !strings.Contains(resp.Result, "already exists") {
```
kraney: This doesn't use strings.ToLower to be case insensitive like the case just below. I don't know if case is ever a factor for this message, though, so it's a minor point.

kraney (approved):
This looks good if the case sensitivity comment is not a concern.

topo/node/juniper/juniper.go
manizzzz: Updated to use strings.ToLower for case-insensitive matching.

